PR #61, opened 2015-01-23
Improve coverage

mstone121:


vegitron:
There's a conflict now in the pull request.  Can you merge from upstream?  I'd expect the conflict to be minimal - probably things added in both forks in the same spot, both of which should be kept.  Let me know if you need a hand.

PR #137, opened 2015-05-21
Feature/course eval

mstone121:
Changed name of BILL AVERAGE TEACHER to Nancy O'Brien-Abël


vegitron:
Unit tests aren't passing.  You should be able to run the tests on your build - let me know if there are any problems doing so before the next pull request.

PR #205, opened 2015-10-06
Resolve MUWM-3390

vegitron:


vegitron:
Doing a combo of this + logout instead.

PR #345, opened 2016-06-03
D/feature/muwm 3677 libapi

mattventura:
- Adds some library subject guide assertions
- Fixes non-functional assertion (`self.assertRaises(KeyError, section.get("canvas_url"))`)
- Moves repeated code into methods
- Moves skipIf to the class level
- Adds jbothell test


mattventura:
Replaced with #347

PR #392, opened 2016-10-19
Moving from requirements.txt to setup.py

vegitron:


vegitron:
@fanglinfang no changes required - requirements.txt now has the content needed to run setup.py with the proper arguments:
https://github.com/uw-it-aca/myuw/pull/392/files#diff-b4ef698db8ca845e5845c4618278f29aR1

with that, does this look good?

PR #398, opened 2016-10-31
Fang/1

fanglinfang:


myuw/apps.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import unicode_literals
```
vegitron: What does this file get us?  It didn't look like this was required.

setup.py
```diff
@@ -23,19 +23,19 @@
...
         'BeautifulSoup',
         'urllib3==1.10.2',
         'django-templatetag-handlebars',
-        'Django-UserService==1.0.2',
+        'django-userservice==1.2.1',
         'unittest2',
         'AuthZ-Group',
         'python-binary-memcached',
         'UW-RestClients',
+        'PermissionsLogging',
```
vegitron: This isn't an app requirement, right?  This is a project level requirement.  On our hosts that's managed through a separate ansible component.

vegitron: This isn't an app requirement, right?  This is a project level requirement.  On our hosts that's managed through a separate ansible component.

PR #405, opened 2016-11-08
Mikes/feature instructor schedule

mikeseibel:


myuw/dao/exceptions.py
```diff
@@ -6,3 +6,8 @@
 class EmailServiceUrlException(Exception):
     """Unhandled email domain or malformed email address"""
     pass
+
+
+class UnsupportedAffiliationException(Exception):
```
vegitron: This exception doesn't seem to be raised anywhere, but it's tested for when testing is_instructor.  

The other is_X checks return true/false though, rather than raising exceptions.  I'm not sure where this would be used, if staying consistent with the older code

PR #416, opened 2016-11-30
Bugfix/choose new version

fanglinfang:
1. Enable users to choose new and old version via a link (added a new dao.user module, updated views.choose, added 2 new unit tests).
2. Refactored unit tests: clean up duplicated code to get a request with a specific user or with a specific date.


myuw/management/commands/change_pref.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument('netid')
+        parser.add_argument('pref', choices=['old', 'new'])
+
+    def handle(self, *args, **options):
+        netid = options['netid']
+        pref = options['pref']
+
+        if pref == 'old':
```
fanglinfang: i am the person who have handled the switch as users couldn't do it themselves.

myuw/views/choose.py
```diff
@@ -1,29 +1,19 @@
...
-    obj.use_legacy_site = False
-    obj.save()
-
+    set_preference_to_new_myuw(get_netid_of_current_user())
     return HttpResponseRedirect(reverse("myuw_home"))
 
 
 @login_required
 def old_site(request):
-    if "POST" == request.method:
```
fanglinfang: As the new_site method in the same file couldn't enforce POST and the expected usage of this link would be low, we have reached an agreement of going without the POST.

myuw/management/commands/change_pref.py
fanglinfang: we've reached the agreement of not needing the admin tool for this. 
A web page for users to choose their preference can be added in the future if it is desired.

PR #419, opened 2016-12-15
adds instructor schedule future and past api

mikeseibel:


myuw/dao/instructor_schedule.py
```diff
@@ -71,6 +72,30 @@ def get_current_quarter_instructor_schedule(request):
...
+
+
+def get_past_quarter_instructor_schedule(request):
+    """
+    Return sections instructor is teaching in a past quarter
+    """
+    if get_next_quarter(request) == get_current_quarter(request):
+        return None
+    term = get_current_quarter(request)
+    offset = int(request.GET.get('offset', 1))
```
vegitron: Same issue here

myuw/views/api/instructor_schedule.py
```diff
@@ -157,3 +158,55 @@ def GET(self, request):
                                        request)
         except Exception:
             return handle_exception(logger, timer, traceback)
+
+
+class InstScheFutuQuar(InstSche):
+    """
+    Performs actions on resource at /api/v1/instructor/schedule/next/.
```
vegitron: This comment doesn't match the format in urls.py

myuw/dao/instructor_schedule.py
```diff
@@ -71,6 +72,30 @@ def get_current_quarter_instructor_schedule(request):
     return schedule
 
 
+def get_future_quarter_instructor_schedule(request):
```
vegitron: Our approach with non-current quarters in the DAO is to pass in a term:

https://github.com/uw-it-aca/myuw/blob/master/myuw/dao/schedule.py#L40

myuw/views/api/instructor_schedule.py
```diff
@@ -157,3 +158,55 @@ def GET(self, request):
...
+        """
+        GET returns 200 with a future quarter course section schedule
+        Integer "offset" parameter sets how many quarters forward (default: 1)
+        @return class schedule data in json format
+                status 404: no schedule found (not registered)
+                status 543: data error
+        """
+        timer = Timer()
+        term = get_current_quarter(request)
+        offset = int(request.GET.get('offset', 1))
```
vegitron: We also specify the term in the url, like https://github.com/uw-it-aca/myuw/blob/master/myuw/views/api/future_schedule.py#L20

PR #424, opened 2016-12-21
Feature/speed testing

vegitron:


myuw/dao/term.py
```diff
@@ -318,3 +322,35 @@ def get_eod_specific_quarter_last_instruction(year, quarter):
...
+
+# The affilliation method caches values on the request object, but this one
+# is just designed to get values into our caching system.
+def _get_term_method(year, quarter):
+    def generated(request):
+        get_specific_term(year, quarter)
+    return generated
+
+
+def current_terms_prefetch(request):
```
fanglinfang: May consider adding pre-fetch of get_next_quarter.

vegitron: I expected that to be needed, but we get the terms around the current term by year/quarter.  That's why it does this work: methods.append(_get_term_method(year, quarter))

PR #426, opened 2016-12-22
adds instructor future/past nav, cards

mikeseibel:


myuw/dao/term.py
```diff
@@ -148,6 +148,13 @@ def is_past(term, request):
     return term.get_eod_last_final_exam() < get_comparison_datetime(request)
 
 
+def is_future(term, request):
+    """
+    return true if the given term is in the future
+    """
+    return term.get_bod_first_day() > get_comparison_datetime(request)
```
vegitron: If it's after the grade submission deadline, but before the first day of the next quarter, the next quarter is the "current" quarter.  You can see the logic in https://github.com/uw-it-aca/myuw/blob/master/myuw/dao/term.py#L125 , with the extra check in get_current_quarter.   Does that detail matter?

vegitron: Despite that, summer A switches to current term in the front-end on the 24th of june, which is the first day of the term. Looks like instructor and student are doing the same thing, so i'm switching over to approve.

PR #434, opened 2016-12-30
Bug/muwm 3774

charlon:
Final work on 500 error page.


myuw/templates/500.html
```diff
@@ -101,53 +150,62 @@
...
+        <div id="app_footer">
+
+            <div class="footer-bar" role="contentinfo">
+                <div class="container footer-links" style="position:relative;">
+                    <ul>
+                        <li><a href="https://myuw.washington.edu/cgi-bin/contactmyuw.cgi"><i class="icon-envelope"></i> Contact</a></li>
+                         <li><a href="http://www.washington.edu/online/terms/">Terms</a></li>
+                         <li style="border:none;"><a href="http://www.washington.edu/online/privacy/">Privacy</a></li>
+                    </ul>
+                    <div>&copy; {% now "Y" %} University of Washington</div>
```
vegitron: The {% now %} won't work when statically rendered during a 500.

charlon: Ahhh! Good catch. I had removed all the {% static %} stuff, but missed this one.

charlon:
The latest commit removes the card class.

PR #435, opened 2017-01-03
Fixes textbook card incorrect sizing. Fixes MUWM-3542

charlon:
This is a simple CSS change for the #TextbookCard id. The "h3" was overriding the default card h3 styles - which should not be the case. Removing the override styles fixes the bug and sets the h3 back to the correct default sizing. I verified that the h3 size was (and should be) the same for every card header.


charlon:
Closing... should not be trying to merge w/ Master branch.

PR #437, opened 2017-01-05
Resolves MUWM-3789

devights:


charlon:
test comment

devights:
@charlon That was never an issue on the textbook card, those already handle the bookstore going down correctly in develop

@vegitron that's reasonable, PR amended

PR #462, opened 2017-01-27
Fang/dao mailman

fanglinfang:


myuw/dao/mailman.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    person = get_person_of_current_user()
+    schedule = _get_instructor_schedule(person, term)
+    json_data = {
+        "year": term.year,
+        "quarter": term.quarter,
+        }
+    json_data["email_lists"] = []
+    for section in schedule.sections:
+        json_data["email_lists"].append(
```
vegitron: There are three layers of sequential work here that would be good to put in parallel.

There's each section
There's primary and secondary sections
There's each secondary section

Those should probably be threaded

myuw/dao/mailman.py
```diff
@@ -0,0 +1,109 @@
...
+                                url, flags=re.IGNORECASE)
+            section = _get_section(primary_section.term,
+                                   primary_section.curriculum_abbr,
+                                   primary_section.course_number,
+                                   section_id)
+            secondaries.append(get_single_email_list_by_section(section))
+    return secondaries
+
+
+def email_list_prefetch():
```
vegitron: My hunch is prefetch isn't going to be helpful here.  Can we walk through a timeline of the api calls?

myuw/dao/mailman.py
vegitron: somewhat related, what happens for the users with very many sections or secondary sections?

myuw/dao/mailman.py
```diff
@@ -0,0 +1,233 @@
...
+from myuw.dao import get_netid_of_current_user
+from myuw.dao.instructor_schedule import get_instructor_schedule_by_term
+
+
+logger = logging.getLogger(__name__)
+MAILMAN_ADMIN_URL = "https://mailman.u.washington.edu/mailman/admin/%s"
+section_id_ext_pattern = r'.*/course/\d{4},[^/]+/([A-Z][A-Z0-9]?).json$'
+
+
+def get_admin_url(list_name):
```
vegitron: Should this be in restclients?

myuw/dao/mailman.py
```diff
@@ -0,0 +1,233 @@
...
+def get_section_id(url):
+    return re.sub(section_id_ext_pattern, r'\1',
+                  url, flags=re.IGNORECASE)
+
+
+def get_all_secondary_section_lists(primary_section):
+    secondaries = []
+    if primary_section.linked_section_urls:
+
+        if len(primary_section.linked_section_urls) > 1:
```
vegitron: I'd rather have just one block of code, even if it leads to an unneeded thread.

myuw/dao/mailman.py
```diff
@@ -0,0 +1,233 @@
...
+    schedule = get_instructor_schedule_by_term(term)
+
+    if len(schedule.sections) >= 0:
+        json_data["instructor_term_list"] =\
+            get_instructor_term_list(get_netid_of_current_user(),
+                                     term.quarter,
+                                     term.year)
+
+    json_data["email_lists"] = []
+    if len(schedule.sections) == 1:
```
vegitron: Same here - I'd like to avoid the extra case in code.

myuw/dao/mailman.py
```diff
@@ -0,0 +1,233 @@
...
+        json_data["email_lists"].append(
+            get_section_email_lists(schedule.sections[0],
+                                    include_secondaries_in_primary))
+
+    if len(schedule.sections) > 1:
```
vegitron: I'm wondering what happens with > 20 or so sections.  I don't want to block this until the design of large number of sections is done, but I also don't want this to crash myuw if we're stuck making a bunch of requests.  Can you get sign-off for just returning nothing if there's more than some number of sections?

myuw/dao/mailman.py
fanglinfang: yes.

myuw/dao/mailman.py
fanglinfang: ok

myuw/dao/mailman.py
fanglinfang: ok

myuw/dao/mailman.py
fanglinfang: that would be up to instructor_schedule.py. 
do we need to make a separate decision for mail list?

vegitron: That's a good question.  While we're waiting to see how all of the section related code needs to change, I'd rather err on the side of failing intentionally, rather than trying to do too much.  

I guess I'm not really viewing this as making a decision, but more like building in an escape valve

fanglinfang: At this point i need to see 20+ secondary sections displayed on the request email list page.

PR #466, opened 2017-01-27
adds final exam confirmation link

mikeseibel:


myuw/templates/handlebars/card/instructor_schedule/course_sche_panel.html
```diff
@@ -55,7 +55,7 @@ <h5 class="scheduleHeading">
 </table>
 
 <h5 class="scheduleHeading">
-FINAL EXAM
+  FINAL EXAM {{#if final_exam.no_exam_or_nontraditional}}{{else}}{{#if final_exam.is_confirmed}}{{else}}(<a href="https://sdb.admin.uw.edu/sisMyUWClass/uwnetid/{{netid}}/finalexam.asp?{{capitalizeString quarter}}+{{year}}&sln={{sln}}" target="_blank">Confirm</a>){{/if}}{{/if}}
```
vegitron: Is there going to be something in the first #if block, or should that be moved to unless?

fanglinfang: i see this as a work around due to missing the object method not_confirmed.

vegitron: I didn't actually notice the second block that was just in an {{else}}.  I was suggesting {{#unless final_exam.no_exam_or_nontraditional}}

mikeseibel: since there wasn't a spec, i was just leaving it open, but unless for now is fine

PR #469, opened 2017-01-31
Bug/muwm 3571

wmwash:


vegitron:
Events and Activities doesn't load, but that doesn't seem critical

PR #472, opened 2017-01-31
scales to many sections on teaching page

mikeseibel:


mikeseibel:
the handlebars partial is extraneous to the scaling work. a little section page prep that slipped in.

mikeseibel:
done.  i was testing by setting max sections to one

PR #473, opened 2017-02-01
update email link on course card

fanglinfang:


myuw/views/api/instructor_schedule.py
```diff
@@ -78,7 +79,14 @@ def load_schedule(request, schedule, summer_term=""):
             section_data["lib_subj_guide"] =\
                 get_subject_guide_by_section(section)
         except Exception as ex:
-            logger.error(ex)
+            logger.error("get lib_subj_guide: %s" % ex)
+            pass
+
+        try:
+            section_data["email_list"] = get_section_email_lists(
```
vegitron: Below this (on line 105) there's a thread that's started per section.  Can this be moved into that thread?

fanglinfang: would you like get_subject_guide_by_section also moved?

PR #482, opened 2017-02-09
class website error feedback, add website popup, course resource indi…

mikeseibel:
…rect

biggest change is slicing up the course card resources into partials to minimize source collisions moving forward

### DIFF
--- a/myuw/dao/class_website.py
+++ b/myuw/dao/class_website.py
@@ -35,14 +35,8 @@ def get_page_title_from_url(url):
         soup = BeautifulSoup(html)
         return soup.title.string
     except DataFailureException as ex:
+        raise
+    except Exception as ex:
         logger.error("get_page_title_from_url(%s)==>%s" % (url, ex))
 
-
-def is_live_url(url):
-    try:
-        _fetch_url('HEAD', url)
-        return True
-    except DataFailureException:
-        pass
-
-    return False
+    return None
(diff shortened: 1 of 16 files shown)


myuw/dao/class_website.py
```diff
@@ -35,14 +35,8 @@ def get_page_title_from_url(url):
         soup = BeautifulSoup(html)
         return soup.title.string
     except DataFailureException as ex:
+        raise
+    except Exception as ex:
```
vegitron: pages without a title will raise an attributeerror.  i think we'd want to avoid logging that

myuw/views/api/instructor_schedule.py
```diff
@@ -68,8 +83,8 @@ def set_course_resources(section_data, section):
 
     if section.class_website_url:
         t = ThreadWithResponse(
```
vegitron: If you're going to use threadwithresponse, I don't think it should track the response and also modify a passed in dictionary.  i'd be fine w/ either a more complex value returned by the function, or moving away from threadwithresponse

myuw/dao/class_website.py
```diff
@@ -33,16 +33,11 @@ def get_page_title_from_url(url):
     try:
         html = _fetch_url('GET', url)
         soup = BeautifulSoup(html)
-        return soup.title.string
+        if hasattr(soup, 'title'):
```
vegitron: >>> from BeautifulSoup import BeautifulSoup
>>> soup = BeautifulSoup("<html>no title</html>")
>>> print hasattr(soup, 'title')
True
>>> print soup.title.string
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'string'
>>>

PR #483, opened 2017-02-10
class web page error ui, event tidying

mikeseibel:


myuw/dao/class_website.py
```diff
@@ -19,9 +20,15 @@ def _fetch_url(method, url):
         logger.error("_get_html_from_url(%s)==>%s" % (url, ex))
         return None
 
-    pool = get_con_pool("%s://%s" % (p.scheme, p.netloc), socket_timeout=2)
-    response = get_live_url(
-        pool, method, p.hostname, url, {'ACCEPT': 'text/html'})
+    headers = {'ACCEPT': 'text/html'}
+    if p.netloc == 'localhost':
```
vegitron: I'd rather see this switch happen w/ settings, rather than hostname.  If not, the check you put in for security below - replacing '..' with '' should be specifically tested.

mikeseibel: the thought was to minimize config and figured since "localhost" has a pretty specific meaning so it's clear what what to expect when you see it in the mock resource.

if it's in config, what's the default?  are you suggesting mocked resources with links to mock websites get changed too?

vegitron: I think it's fairly clear, though there might be some resistance to it as the drive for "more realistic" mock data moves forward.  

I'd say the default would go to mock resources, to keep it in parity with the other live/mock resources.  I don't have a strong opinion about what urls are in mock data, but you might be able to avoid some questions if the links go to a working website.

If you want to keep the localhost solution, I'm not too worried about it as long as the security concerns are specifically defined and tested.

mikeseibel: how's the next commit for a compromise?

PR #485, opened 2017-02-15
Vegitron/photo list

vegitron:


myuw/dao/enrollment.py
```diff
@@ -9,6 +9,18 @@
...
+CLASS_CODES = {
+    "FRESHMAN": 1,
+    "SOPHMORE": 2,
+    "JUNIOR": 3,
+    "SENIOR": 4,
+    "GRADUATE": 5,
+}
+
+
+DEFAULT_CLASS_CODE = 6
```
fanglinfang: What 6 stands for? Are these codes used in SDB?

myuw/views/api/instructor_schedule.py
```diff
@@ -322,3 +326,120 @@ def GET(self, request, year, quarter, curriculum,
...
+        timer = Timer()
+        try:
+            return self.make_http_resp(timer, year, quarter, curriculum,
+                                       course_number, course_section,
+                                       request)
+        except Exception:
+            return handle_exception(logger, timer, traceback)
+
+
+def not_instructor_error():
```
fanglinfang: please move this into rest_dispatch.py

myuw/dao/enrollment.py
vegitron: Those are just numerical values we can use for sorting.  6 is just bigger than the highest value in the dictionary above.  If we add more defined sort orders, that value would need to change.  I could change it to len(CLASS_CODES.keys()+1, if that would help

myuw/views/api/instructor_schedule.py
vegitron: 👍

PR #496, opened 2017-02-17
Mikes/teaching page

mikeseibel:


myuw/views/api/textbook.py
```diff
@@ -34,27 +35,52 @@ def GET(self, request, year, quarter, summer_term):
...
+            if schedule:
+                if summer_term is not None and len(summer_term) > 0:
+                    summer_term = summer_term.replace(",", "")
+                    filter_schedule_sections_by_summer_term(
+                        schedule, summer_term)
 
-            if summer_term is not None and len(summer_term) > 0:
-                summer_term = summer_term.replace(",", "")
-                filter_schedule_sections_by_summer_term(schedule, summer_term)
+                if len(schedule.sections) == 0:
```
vegitron: If someone's an instructor and student, signed up for summer B term, it looks like if this is loaded during A term they'll get data_not_found, before it looks for their instructor book list.

mikeseibel: you might have a point...

PR #504, opened 2017-02-23
Thatssad/teaching page

klpittman:
Some changes to the Teaching page title and selector.
@mikeseibel I also updated the display for long lists of courses. It seems to be working but it would be good if you had a look at the code. Thanks!

### DIFF
--- a/myuw/templates/handlebars/card/instructor_schedule/course_cards.html
+++ b/myuw/templates/handlebars/card/instructor_schedule/course_cards.html
@@ -34,15 +34,21 @@
     </div>
   </div>
   {{/each}}
+
+  {{#if section_references}}
+  <div class="card">
+      <ul class="unstyled-list">
   {{#each section_references}}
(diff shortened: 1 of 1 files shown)


mikeseibel:
What do you think about adding the css below to limit title length?

diff --git a/myuw/static/css/mobile.less b/myuw/static/css/mobile.less
index 07d7161..f98d129 100644
--- a/myuw/static/css/mobile.less
+++ b/myuw/static/css/mobile.less
@@ -1188,6 +1188,15 @@ body {
                     }
                 }
             }
+            .course_website {
+                display: inline-block;
+                white-space: nowrap;
+                overflow: hidden;
+                text-overflow: ellipsis;
+                max-width: 350px;
+                line-height: 1.2em;
+            }
+
         }
         #teaching_resources {

klpittman:
@mikeseibel I think your comment is about limiting the length of the website titles. That's a good idea.

This commit is updating the page title "Teaching" and removing the div cards that were used for each listed course.

PR #507, opened 2017-02-24
class list table view and minor updates

bexar:


klpittman:
HTML sections look good to me. Not sure I have much to contribute for the .py files.

PR #508, opened 2017-02-24
Thatssad/teaching

klpittman:


bexar:
Looks good for desktop! On mobile we'll need to adjust with max-width but that will be part of adjusting the whole layout.

PR #522, opened 2017-03-07
adds logic to use campus affiliation to determine best iasystem link …

klpittman:
…to show in teaching resources.

If possible, show the campus specific link to iasystem. If no affiliation is available, show a link to to the general help page website (let's the user choose which URL they need there).


vegitron:
If the only thing that's different is the link, you might want to only have the if blocks around the opening <a> tag, or even just the url.

klpittman:
@vegitron thanks for the review. Adjusted template to accommodate.

PR #534, opened 2017-03-15
Adding a test for poorly formed br tags

vegitron:
This looks finds &lt;/br&gt; , which is necessary because browsers will render it as a proper &lt;br&gt;, our template precompiler will strip it out - so formatting is different on production than on local builds.

Are there other tags like that which should be disallowed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,8 @@ script:
   - ls -l ./node_modules/.bin/_mocha
   - ls -l ./node_modules/mocha/bin/_mocha
   - pep8 myuw/ --exclude=migrations,myuw/south_migrations,myuw/static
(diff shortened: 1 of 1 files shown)


klpittman:
@vegitron Hmm, I suppose the possibilities are fairly endless. I haven't explored the templates enough to know what the common errors are. 

In the same spirit as the <br> tag, there's no `</input>` tag. Inputs are self-closing. 

Here are more self-closing tags:

- <area />
- <base />
- <br />
- <col />
- <command />
- <embed />
- <hr />
- 
- <input />
- <keygen />
- <link />
- <meta />
- <param />
- <source />
- <track />
- <wbr />

PR #539, opened 2017-03-16
Feature/teaching page grading

mikeseibel:


myuw/static/js/test/global.js
```diff
@@ -0,0 +1,119 @@
...
+    },
+    ajax_stub_restore: function () {
+        if (Environment._stub) {
+            Environment._stub.restore();
+        }
+    },
+    _load_script(script) {
+        $.each(require(script), function(k, v) { global[k] = v; });
+    },
+    _read_template(template_file) {
```
vegitron: This looks like it should get a test

myuw/static/js/test/global.js
```diff
@@ -0,0 +1,119 @@
...
+            .replace(/{\%[ ]*endtplhandlebars[ ]*\%}/, 
+                     '</script>');
+        return template;
+    },
+    _load_template(template_file) {
+        var template = Environment._read_template(template_file);
+
+        while (true) {
+            // pull in server-side includes
+            var m = template.match(/{\%[ ]*include[ ]+["]?([^ \%"]+)["]?[ ]*\%}/);
```
vegitron: This also looks like it should get a test

myuw/static/js/test/global.js
```diff
@@ -0,0 +1,119 @@
...
+        var json_data = JSON.parse(fs.readFileSync(json_file));
+        Environment._stub = sinon.stub($, 'ajax');
+        Environment._stub.yieldsTo('success', json_data);
+    },
+    ajax_stub_restore: function () {
+        if (Environment._stub) {
+            Environment._stub.restore();
+        }
+    },
+    _load_script(script) {
```
vegitron: What do you think about adding a check here to see if there's already a matching name in global?  It probably wouldn't come up often, but if the test suite could say something like 'Unable to load CardFoo from js/card/not_foo.js, already loaded from js/card/foo.js' it might save someone some grief

myuw/static/js/test/global.js
mikeseibel: heh, just looking at that

PR #541, opened 2017-03-16
changing emaillist modal error handling

devights:


klpittman:
Left a screengrab on the trello ticket. Wondering if we can remove the "Create Mailing Lists" button when there's an error like this? thanks!

PR #548, opened 2017-03-28
Feature/upass status

fanglinfang:
1. Upass content
2. https://jira.cac.washington.edu/browse/MUWM-3803


myuw/static/js/card/upass.js
```diff
@@ -0,0 +1,36 @@
...
+    dom_target: undefined,
+
+    render_init: function () {
+        UPassCard.dom_target = $('#UPassCard');
+        WSData.fetch_upass_data(UPassCard.render_upon_data, UPassCard.render_error);
+    },
+
+    render_error: function (status) {
+        if (status === 404) {
+            UPassCard.dom_target.hide();
```
devights: Is hiding on 404 the appropriate behavior?  From what I saw the upass API shouldn't return a 404 so it might be valuable to show the error card when the API does 404.

fanglinfang: Thanks! I revised

PR #572, opened 2017-04-18
Devights/site search

devights:


devights:
To review this you'll need the new settings key:
`GOOGLE_SEARCH_KEY = 'your search key'`

Also, by moving to use our own custom form we don't need to deal with that redirect BS so it'll make development easier as search will work without much config!

PR #577, opened 2017-04-21
Feature/course card display closed section

hlee2:
fix span tag


fanglinfang:
Not sure why we want to merge into develop.
anything merging to develop is ready to be released, which means we have voted yes in go-no-go.

PR #589, opened 2017-04-27
Feature/banner message

devights:


myuw/models/__init__.py
```diff
@@ -136,3 +137,33 @@ class Meta:
 class UserMigrationPreference(models.Model):
```
fanglinfang: could you move BannerMessage into its own model file after release?

myuw/static/js/banner/messages.js
```diff
@@ -0,0 +1,23 @@
+var MessageBanner = {
+    dom_target: undefined,
+
+    render_init: function(dom_taget) {
+        MessageBanner.dom_target  = dom_taget;
+        WSData.fetch_message_data(MessageBanner.render);
+    },
+
+    render: function () {
+        var message_data = WSData.message_data();
```
fanglinfang: Would you hide the target when message length is zero?

PR #602, opened 2017-05-01
Feature/pending major

ezturner:


myuw/dao/enrollment.py
```diff
@@ -77,6 +79,81 @@ def get_main_campus(request):
...
+    """
+    Takes in two lists of degrees (either major or minors) and checks to see
+    if they are the same. Returns True if so, False if they are different
+    """
+    if len(first) != len(second):
+        return False
+
+    degrees = {}
+    for entry in first:
+        degrees[entry.full_name] = entry
```
fanglinfang: Major and Minor has __eq__ method build in now, which can may your code cleaner

myuw/static/js/banner/profile.js
```diff
@@ -2,6 +2,7 @@ var Profile = {
     render_upon_data: function() {
         var source = $("#profile-content").html();
         var template = Handlebars.compile(source);
+        var profile_data = WSData.profile_data();
         $("#profile").html(template(WSData.profile_data()));
```
fanglinfang: the var on line 5 can be used in line 6

myuw/dao/enrollment.py
```diff
@@ -77,6 +79,81 @@ def get_main_campus(request):
...
+    """
+    if len(first) != len(second):
+        return False
+
+    degrees = {}
+    for entry in first:
+        degrees[entry] = entry
+
+    for entry in second:
+        if entry not in degrees:
```
fanglinfang: 144 -146 seems unnessary.
why not just "if entry not in first"?

PR #617, opened 2017-05-11
Vegitron/quicklinks

vegitron:


klpittman:
@bexar Some small suggestions. 

1. Left alignment on the intro text
myuw_teaching

2. Also, it seems like there will be very few cases/times when someone would see the intro text since it will go away as soon as a link is clicked. Is anything in there important enough to show even when links are shown? Just a thought...

3. Bold text feels more attention grabbing than needed

I wonder if we could just use the same header style as "Recently Visited"? Could just be "More" or something else short?

myuw_teaching

klpittman:
@bexar also, do we have a standard style to use for that error message?

PR #624, opened 2017-05-12
adds js tests for account and profile cards

mikeseibel:


myuw/static/js/test/account_card.js
```diff
@@ -0,0 +1,26 @@
...
+                    "myuw/static/js/card/accounts/account_card.js"
+                ],
+                templates: [
+                    'myuw/templates/handlebars/card/accounts/account_card.html'
+                ]
+            });
+
+            AccountsCard.dom_target = $('#' + render_id);
+        });
+        it("Should render instructor card", function() {
```
vegitron: It would be good to test the inverse as well, to make sure it isn't showing up for non-instructors

PR #626, opened 2017-05-12
Feature/accounts page

mikeseibel:


myuw/static/js/test/account_card.js
```diff
@@ -0,0 +1,26 @@
...
+                    "myuw/static/js/card/accounts/account_card.js"
+                ],
+                templates: [
+                    'myuw/templates/handlebars/card/accounts/account_card.html'
+                ]
+            });
+
+            AccountsCard.dom_target = $('#' + render_id);
+        });
+        it("Should render instructor card", function() {
```
vegitron: what are the other test cases here?  this is one of the happy paths, but there should probably also be a test to make sure people are being rejected properly as well.  it might be easier to test those cases if the user validation code was a separate function

myuw/static/js/handlebars-helpers.js
```diff
@@ -425,3 +425,16 @@ Handlebars.registerHelper('slugify', function(value) {
     var slug = value.replace(/[^\w\s]+/gi, '').replace(/ +/gi, '-');
     return slug.toLowerCase();
 });
+
+Handlebars.registerHelper('phone_number', function(value) {
```
vegitron: Can you add some tests to directly test this?  I see there's an indirect test below, but if there are multiple input formats it would be good to make sure they're all being parsed.  And probably good for a fall-through test, just to be complete.

PR #627, opened 2017-05-16
Hlee2/accounts page

hlee2:
Merging the first design iteration to the feature branch.


klpittman:
The NetID account card on the Accounts page should continue to be shown for all users (assuming they all have NetIDs). It looks like it is being hidden for Students. 

Test with: 

> jbothell

klpittman:
All of the Instructor/Staff cards and work look like they are accounted for here. Good to go on that front.

myuw/static/js/page/accounts.js
```diff
@@ -17,9 +17,9 @@ var RenderAfterAjaxError = function () {
 
 var _render_cards = function(has_medicine) {
     // will ultimately have more than acct and med acct cards here
-    var card_order = [AccountsCard];
+    var card_order = [AccountsCard, HRPayrollCard];
     if(has_medicine){
-        card_order = [MedicineAccountsCard, AccountsCard];
+        card_order.splice(0, 0, MedicineAccountsCard);
```
vegitron: Can you do unshift instead of split?

PR #641, opened 2017-05-23
Feature/boilerplate header

charlon:
New boilerplate work that includes new base layout and improved mobile UI built.


myuw/static/js/landing.js
```diff
@@ -82,6 +83,7 @@ var Landing = {
     _load_mobile_cards: function() {
         Landing._reset_content_divs();
         var mobile_cards = [
+            LibraryCardMini,
```
devights: this should be removed

charlon: okay it has been removed

vegitron:
@klpittman i saw that too, and put together pull request #642 that fixes it and tests to make sure it won't happen again.  it shouldn't block this pull request.

PR #644, opened 2017-05-24
adds profile contextual links

mikeseibel:


myuw/templates/handlebars/card/profile/help_links_card.html
```diff
@@ -0,0 +1,20 @@
+{% load templatetag_handlebars %}
+{% tplhandlebars "profile_help_links_card" %}
+<div class="profile-links" style="position:relative;">
+  <h4>Related Help</h4>
+  <ul class="help_links employee_help_links">
```
klpittman: @mikeseibel Do we need "employee_help_links" here as a class? The list should serve both employees and students, right?

mikeseibel: good eye. that was left over from the first iteration with the two sections

PR #666, opened 2017-06-02
Feature/academics page

fanglinfang:


myuw/static/js/textbooks.js
```diff
@@ -44,7 +49,7 @@ var TextBooks = {
                 color_id: section.color_id,
                 sln: section.sln,
                 books: book_data ? book_data[section.sln] : [],
-                has_books: book_data ? (book_data[section.sln].length > 0) : false,
```
fanglinfang: js error .length on undefined object

vegitron: that's fixed in #665

fanglinfang: too bad we met at merge

PR #669, opened 2017-06-02
Feature/account summaries

vegitron:


bexar:
@charlon can we break the date and the 'week x of n' string into different styles and maybe placement? here's a couple ideas
screen shot 2017-06-02 at 3 43 44 pm
screen shot 2017-06-02 at 3 44 20 pm

PR #684, opened 2017-06-07
Feature/quicklinks-b to develop

bexar:


klpittman:
Getting an error when trying to save custom links:

screen shot 2017-06-07 at 1 24 39 pm

Console:

> Not Found: /api/v1/link
> [07/Jun/2017 13:24:38] "POST /api/v1/link HTTP/1.1" 404 14

klpittman:
On some links, it looks like the "+" icon falls to the next line.

myuw__home

myuw/templates/handlebars/card/quicklinks.html
```diff
@@ -21,56 +21,53 @@
...
             <div class="form-group">
-            
-            <input class="form-control" type="hidden" id="custom-link-edit-id"> 
-            URL: <input class="form-control" type="url" id="custom-link-edit-url"> 
-            Label: <input class="form-control" type="text" id="custom-link-edit-label"> <span id="edit-label-required" style="display: none;">* Required</span>
-            <div class="text-right" style="padding: 10px 0;"><a href='#' id="close-custom-link-edit">cancel</a> <button class="btn btn-primary" type="button" id="quicklinks-save-edits">Save</button></div>
+                <input class="form-control" type="hidden" id="custom-link-edit-id"> 
+                URL: <input class="form-control" type="url" id="custom-link-edit-url"> 
+                Label: <input class="form-control" type="text" id="custom-link-edit-label"> <span id="edit-label-required" style="display: none;">* Required</span>
+                <div class="text-right" style="padding: 10px 0;"><a href='#' id="close-custom-link-edit">cancel</a> <button class="btn btn-primary" type="button" id="quicklinks-save-edits">Save</button></div>
```
klpittman: Should "cancel" be capitalized?

klpittman:
Edit form falls below the next card. 

screen shot 2017-06-07 at 1 30 04 pm

klpittman:
Should we standardize the background color for popin forms like this? 

myuw__teaching

bexar:
@klpittman updates and fixes should all now be committed on feature branch, let me know if there's still something funky

wmwash:
The functionality looks good to me. One content suggestion, and a few visual design tweaks you can make if time allows:

**Recently Visited** - The "Save your recently visited links to your Quick Links...." Feels both hierarchically inaccurate (the whole card is called Quicklinks), and I wonder if it's necessary to have that statement there...I know we moved it from under the header to below the links, but it still feels in the way! ;-) I wonder if we can't just rely on people discovering that feature?   
**Popular Links font size** - is there a reason the popular links are smaller text than everything else in that card? Feels unnecessary to me
**Cancel/Close links** - We can use more space between the link and button; currently a bit cramped.

wmwash:
...and if we do decide to keep that statement, perhaps we can just say, "Save your recently visited links for future access."

bexar:
@wmwash originally the idea of your "Quick Links" was different then recently visited and popular links, which are just provided as suggestions for possible Quick Links, so there was that hierarchical difference and hence the messaging and diff font point size. I think we've slowly moved away from that and I take from what you are suggesting is that that difference is unnecessary. I'm pretty ok with that at this point. The changes are simple so I'll push them before we merge.

wmwash:
Nice iteration, @bexar! Looks great - merge whenever you're ready!

PR #701, opened 2017-06-13
This is the set of features for the big june release

vegitron:


myuw/dao/gws.py
```diff
@@ -78,7 +81,9 @@ class is one of (01, 02, 03, 04, 05, 06),
 
 
 def is_student():
-    return is_undergrad_student() or is_graduate_student()
+    return (is_undergrad_student() or
+            is_graduate_student() or
+            is_pce_student())
```
fanglinfang: this will give PCE students the new version if they access from a mobile device.  should we leave it or change back?

PR #705, opened 2017-06-14
Devights/vs bugfixes

devights:


hlee2:
Are they PCE off-term courses?  I noticed those mock data is slightly off. According to the academic calendar A-term: June 24 -July 24  B-term: July 25 - August 23. TRAIN 101 A  which shows on B-term starts on 7/22, before B-term. 

https://www.washington.edu/students/reg/calendarpast/1213cal.html

PR #711, opened 2017-06-15
Feature/onboarding

wmwash:
Revising banner and modal for onboarding use for new version. Didn't use admin banner tool b/c it doesn't support font awesome (or I couldn't figure out how to do it, anyway). 

Removed as much old modal banner code as I could see to keep code clean.

### DIFF
--- a/myuw/static/css/mobile.less
+++ b/myuw/static/css/mobile.less
@@ -553,19 +553,20 @@ label { font-weight: normal; }
 			}
 		}
 	}
-	.message-dialog {
-		.modal-content {
-			h3 {margin-bottom: 10px;font-family: 'encode_sans_normalbold', sans-serif; font-weight: 600; font-size: 18px;color: #85754d;}
-			.modal-header {border-bottom: none;padding-bottom: 0;}
-			.modal-body {padding: 0 15px;
-				ul { padding-left: 0px;
-					li {list-style: none; padding-bottom: 10px;}
-				}
+}
+.message-dialog {
+	.modal-content {
+		h3 {margin-bottom: 10px;font-family: 'encode_sans_normalbold', sans-serif; font-weight: 600; font-size: 18px;color: #85754d;}
+		.modal-header {border-bottom: none;padding-bottom: 0;}
+		.modal-body {padding: 0 15px;
+			ul {list-style: none;padding-left: 0;
+				li {padding-bottom: 10px;}
 			}
-			.modal-footer {border-top: none;}
 		}
+		.modal-footer {border-top: none;}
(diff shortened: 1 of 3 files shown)


klpittman:
Looks good. We should add the ability to use font-awesome icons in the admin tool for the future.

Not a blocker, but what is the plan for hiding the banner? When will it disappear or will the user ever have the ability to close it?

myuw/static/css/mobile.less
```diff
@@ -553,20 +553,21 @@ label { font-weight: normal; }
...
-			.modal-body {padding: 0 15px;
-				ul {
-					li {line-height: 25px;}
-				}
+}
+.message-dialog {
+	.modal-content {
+		h3 {margin-bottom: 10px;font-family: 'encode_sans_normalbold', sans-serif; font-weight: 600; font-size: 18px;color: #85754d;}
+		.modal-header {border-bottom: none;padding-bottom: 0;}
+		.modal-body {padding: 0 15px;
```
klpittman: Not a blocker, but if we're editing this code anyway we could update the padding/margins to fall into our 8px/16px system.

myuw/templates/base.html
```diff
@@ -114,14 +114,10 @@ <h1 class="sr-only">MyUW</h1>
 
         <!-- message banner (desktop only) -->
         <div class="myuw-banner-desktop" id="app_messages" role="complementary">
-
             <div class="message-banner">
-                {% if user.affiliations.grad %}
                 <div class="message-text">
-                    <h2>Welcome to MyUW!</h2>&nbsp; MyUW shows you personal, critical, and timely information. <a id="modal_toggle" href="#" class="btn btn-link" data-toggle="modal" data-target="#message_modal">Learn how it works.</a>
+                    <h2>New in MyUW</i><i class="fa fa-bullhorn" aria-hidden="true" style="font-size: 20px; padding: 0 6px 0 6px;"></i>&nbsp;</h2>Quick Links, Pages, Search, and more! &nbsp;<a id="modal_toggle" href="#" class="btn btn-link" data-toggle="modal" data-target="#message_modal">Learn what's new in MyUW</a>
```
klpittman: Actually, looks like you have a random closing </i> tag in here you'll want to fix.

`<h2>New in MyUW</i>`

klpittman:
@bexar Sounds like "alert" might be too assertive based on this: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

bexar:
@klpittman ah yeah, makes sense

PR #713, opened 2017-06-16
Feature/accounts card order

devights:
rebasing on qa wasn't straightforward so this is just going to develop

### DIFF
--- a/myuw/static/js/accounts.js
+++ b/myuw/static/js/accounts.js
@@ -28,44 +28,71 @@ var AccountsPage = {
     },
 
     load_cards_for_viewport: function() {
-        if (get_is_desktop()) {
-            AccountsPage._load_desktop_cards();
-        } else {
-            AccountsPage._load_mobile_cards();
+        AccountsPage._load_cards();
+    },
+
+    _load_cards: function(){
+        var cards = AccountsPage._get_card_order_by_affiliation();
+        if(cards){
+            Cards.load_cards_in_order(cards, $("#accounts_content_cards"));
+            AccountsPage.order_card_list();
         }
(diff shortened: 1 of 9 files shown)


klpittman:
@fanglinfang I'm not sure I'm the best person to review this. @hlee2 would be a better judge if the cards are all in the correct order as she's most familiar with the spec.

hlee2:
- As the viewport size increase or decrease, cards gets stacked to a single column around 1280w.
- Stacked cards(1 column) stay even after the viewport size increase to 1280+. Expected to shift to 2 columns.

hlee2:
In 2 column display, the library card and the U-pass card are switched for javerage, jbothell, jnew. But the right order for jinter and jeos. 
Expected card order is 1)Tuition 2)Husky 3) Library 4)Upass 5) NetID from left to right.

devights:
@hlee2 I can't repro your issue with the card order, it's showing in the defined order top>bottom, left>right.

otherwise other issues have been addressed.

vegitron:
The goal of adding the remove_card method was reducing the amount of code, so that it would be easier for card authors and more reliable.  A remove_card method per card just adds more code around what was there before.  I was thinking of something like this, potentially in myuw_m.js:

function remove_card(target) {
    $(target).remove();
    $(window).trigger("card-hide");
}

devights:
@vegitron Got it, I mis-understood what you were asking first and figured there was some benefit to allowing each card to add extra functionality to it's remove_card method (unsure what that would be at this point).  made the method global!

hlee2:
Verified 1) changing between single column and 2 columns as the viewport increase/decrease 2) Card order.

PR #715, opened 2017-06-16
Allowing classes for font-awesome, and a small set of styles

vegitron:
@wmwash i think this gives you the styling control you need.  if not, can you request changes on here w/ anything more?


bexar:
For icons in banner, I would also suggest allowing aria-hidden="true" to hide icons from screen readers

vegitron:
OK, added aria-hidden

PR #718, opened 2017-06-16
Thatssad/teaching visual design

klpittman:
Lots of little tweaks to spacing etc. for the teaching page. Some CSS refactor to remove duplicates. Adjustments for responsive.


bexar:
@klpittman can we bring the link up to the same row as section header when there is no final exam scheduled?

screen shot 2017-06-19 at 8 20 40 am

bexar:
Also, I'm wondering if we could incorporate the course colors we use into the teaching course details somehow. In student course cards for example, we added a colored top border, course type colored text, etc.

klpittman:
> can we bring the link up to the same row as section header when there is no final exam scheduled?

@bexar, The logic there is interesting. It looks like the system thinks there is actually a final exam, but then no dates get printed for the start or finish time. So the line you are seeing is actually the start/end time en-dash. I've been assuming this is just a mock data issue, but perhaps @mikeseibel can confirm? 

It is in the template: myuw/templates/handlebars/card/instructor_schedule/course_sche_panel.html

klpittman:
@bexar I believe I've addressed the final exam schedule confirmation link issue as well. Take a look when you have a chance.

PR #722, opened 2017-06-19
Changing MyUW in header to link to home. Adding Sign Out to footer in…

wmwash:
… mobile. Removing old link/modal used during initial desktop release.


myuw/templates/base.html
```diff
@@ -95,7 +95,7 @@
 
         <div class="myuw-navigation">
             <div class="myuw-wrapper" style="background-image: url({% static 'images/w-logo-white.png' %})">
-                <div class="myuw-title">MyUW</div>
+                <div class="myuw-title"><a href="/home/">MyUW</a></div>
```
klpittman: shouldn't this link destination just be `href="/"`?

PR #735, opened 2017-06-23
Feature/teaching summary schedule

mikeseibel:


klpittman:
screen_shot_2017-06-23_at_2_03_40_pm

Looks like when the "Room to be arranged" msg is there, it is causing some layout issues. Looks like it is from the colspan on that cell. If you're just using the existing template we may need to add some logic in there to handle this differently on the homepage than we do on the Teaching page.

PR #740, opened 2017-06-23
Bugfix/acal page fixes

devights:
resolves MUWM-3842  and MUWM-3843

### DIFF
--- a/myuw/static/js/academic_calendar.js
+++ b/myuw/static/js/academic_calendar.js
@@ -12,9 +12,24 @@ var AcademicCalendar = {
         var source = $("#calendar_events").html();
         var template = Handlebars.compile(source);
         var grouped = AcademicCalendar.group_events_by_term(events);
-        $("#main-content").html(template({terms: grouped}));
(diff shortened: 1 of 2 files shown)


vegitron:
For the MUWM-3842 fix, the root cause was events that don't have a year attached.  This fix prevents the duplicate headers, but if you take the year off independence day 2013 in mock data, it's out of order, and there's a blank header row.
screen shot 2017-06-26 at 9 41 14 am

vegitron:
What's the goal with the fix for MUWM-3843?  That production data has independence day in spring quarter.  Grouping it there doesn't seem like it's the right thing to do.

devights:
I'm waiting on William's opinion on the 3843 issue since he had an issue with the current handling, for the issue with blank headers there's this bug MUWM-3841 that's waiting on ux.

devights:
Closing due to conversation with William, plan will be to fix ical data in trumba, consider logging events without term data in future to identify early and fix (or find a better data provider)

PR #744, opened 2017-06-26
shows notices link for students only

mikeseibel:
test visually with students: javerage, eight, jnew and jpce and non-students bill, billtac


klpittman:
@wmwash Can you take a look at this too. Are there any other affiliations besides students who should see Notices?

PR #753, opened 2017-06-29
Updates employee profile to new 2 column card layout.

charlon:
This styles the employee profile in the same 2-column layout as student profiles.

### DIFF
--- a/myuw/templates/handlebars/card/directory_info.html
+++ b/myuw/templates/handlebars/card/directory_info.html
@@ -1,78 +1,114 @@
 {% load templatetag_handlebars %}
     {% tplhandlebars "directory_info_card" %}
-    <div class="card" data-type="card" data-name="DirectoryInfoCard" data-identifier="{{ title }}">
-        <h3>Employee Profile</h3>
+
+    <div class="myuw-card" data-type="card" data-name="DirectoryInfoCard" data-identifier="{{ title }}">
+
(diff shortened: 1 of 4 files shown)


klpittman:
myuw__profile

@charlon what do you think of these changes?

* Remove full name since it is already above. Student card doesn't have it so it feels strange to see it here, especially when you can see both cards.
* Add "Work" to the "Address" label. Just clarifies which address is shown.

klpittman:
myuw__profile

If we remove that separator between Email and Phone Number, we can think of that as all the profile info, and then perhaps put the "Manage profile in workday" link below it all similar to how the Student Card is handling it. 

@charlon what do you think?

myuw/templates/handlebars/card/directory_info.html
```diff
@@ -59,9 +41,6 @@ <h4 class="myuw-card-row-heading">Email 2</h4>
                 </div>
                 {{/if}}
 
-            </div>
-
-            <div class="myuw-card-section" property="workLocation" typeof="Place">
 
                 <span class="sr-only" property="name">UW Office</span>
```
klpittman: Let's put the `<span class="sr-only">` here inside the `<div class="col-sm-3">` as the first element. Where it is placed now is causing some inconsistent spacing due to how the CSS is written.

PR #755, opened 2017-07-03
Bug/muwm 3851

charlon:
Checked if hfs/library data exists before we display the "myuw-account-summaries" div. The div will not be rendered if it's empty... which removes unwanted empty spaces. Also, updated the css to have a top border instead of bottom border for the account summaries div. Tablet/desktop views should remain the same. 

Verified against users: bill, seagrad, javerage, jnew, jintern, jerror, jbothell


klpittman:
myuw__home

Overall it is looking pretty good. Do we want just a tad more space between these two elements? It feels a little tight there.

wmwash:
All looks good. Agree with Kevin that it needs just a bit more spacing btwn bottom two rows on mobile

PR #764, opened 2017-07-10
Staff homepage hr card

bexar:
Adds HR card to homepage for staff affiliation only.


klpittman:
screen shot 2017-07-10 at 11 50 35 am

Doesn't seem to be working. Am I missing something?

bexar:
@klpittman add 'workday_account_card' to enabled feature list in settings.py

bexar:
@klpittman yes, it should be there eventually but weren't we hiding it for student employees until workday release?

klpittman:
You're right, I had forgotten that we decided to hide it from Student staff because it doesn't make as much sense as designed for them.

myuw/dao/affiliation.py
```diff
@@ -61,6 +61,7 @@ def get_all_affiliations(request):
             "undergrad": is_undergrad_student(),
             "student": is_student(),
             "pce": is_pce_student(),
+            "staff_employee": is_staff_employee(),
```
fanglinfang: You may add it in affiliation_prefetch

myuw/static/js/landing.js
```diff
@@ -41,24 +41,31 @@ var Landing = {
...
-            InternationalStuCard,
-            SummaryScheduleCard,
-            VisualScheduleCard,
-            TextbookCard,
-            FutureSummaryScheduleCard,
-            FutureQuarterCard1,
-            SummerRegStatusCard1
-        ];
+        var desktop_body_cards;
+        if(window.user.staff_employee) {
```
fanglinfang: What if it is a staff employee and also taking a course?

klpittman: Good question. In that case, we should default to the view typical Student would see. Visual schedule on the homepage. Workday card on the Accounts page. 

@bexar feel free to chime here if you think differently or I'm mistaken.

myuw/dao/affiliation.py
mikeseibel: good catch

fanglinfang:
I approved. but it'd better to have a mock user for verifying the last case.

PR #775, opened 2017-07-18
Feature/more acad cals

fanglinfang:


bexar:
@fanglinfang Will adding the calendar card/widget to homepage be on a separate branch?

PR #779, opened 2017-07-19
Thatssad/instructor admin

klpittman:
Minor update to 

- only print the instructor-term-cards container when needed (normal section cards)
- add a title to the admin course card list
- add a note to the admin course card list

Test with "bill100"

### DIFF
--- a/myuw/templates/handlebars/card/instructor_schedule/course_cards.html
+++ b/myuw/templates/handlebars/card/instructor_schedule/course_cards.html
@@ -56,7 +56,7 @@
 
   {{#if section_references}}<!-- Summary list of courses for over 10 -->
   <div class="myuw-card" aria-labelledby="your_courses">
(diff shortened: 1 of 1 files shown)


myuw/templates/handlebars/card/instructor_schedule/course_cards.html
```diff
@@ -51,20 +52,23 @@ <h4 class="myuw-card-row-heading">Enrollment</h4>
       </div>
       {{/each}}
   </div>
+  {{/unless}}
 
-  {{#if section_references}}
-  <div class="card">
+  {{#if section_references}}<!-- Summary list of courses for over 10 -->
+  <div class="myuw-card" aria-labelledby="your_courses">
+      <h3 class="myuw-card-title">{{ucfirst quarter}} {{year}} Courses</h3>
```
bexar: Does the H3 here also need id="your_courses"?

klpittman: good catch

PR #802, opened 2017-07-26
Adds comma separator to majors list. Fixes MUWM-3881

charlon:
Adds comma separator for each item in the list. Uses the "unless" conditional to check for the last item in the list and excludes the comma for that item. Single majors should always get caught by the unless they will be the only (and last?) item. 

I couldn't find where in the mock data we can add additional majors... so i'm hoping this just works.


klpittman:
@charlon you can also use this class in the components.less file:

```

/*** comma delimited list ***/
    .myuw-card-comma-list {
        display: inline;
        list-style-type: none;
        padding-left: 0;
    }
```

PR #810, opened 2017-07-27
Fix future quarter 10+ course issue

fanglinfang:
Personal build
user: bill100
date: 2013-01-15 00:00:01


klpittman:
myuw__teaching

@fanglinfang When I go to the /teaching page I'm still seeing a blank page. Is this expected with this work?

PR #834, opened 2017-08-10
muwm-3930 fixes email url for uw office 365 users

bexar:


bexar:
Adding @fanglinfang to review because of failing tests

PR #837, opened 2017-08-10
muwm-3935 adds enrollment count to summary schedule card current quarter

bexar:


myuw/templates/handlebars/card/summary/schedule.html
```diff
@@ -55,15 +55,9 @@ <h4 class="myuw-card-row-heading" style="display:inline-block; margin-bottom:0px
                                     {{/if}}
                                 {{/unless}}
 
-                                {{#if ../../show_enrollment}}
                                     <td class="course-type">
                                         {{#if ../current_enrollment}}<a target="_blank" href="/teaching/{{ ../../year }},{{ ../../quarter }},{{ ../curriculum_abbr }},{{ ../course_number }}/{{ ../section_id }}/students" class="course_class_list" rel="{{ ../curriculum_abbr }} {{../course_number}} {{../section_id}}">{{../current_enrollment}}</a>{{else}}0{{/if}}&nbsp;of&nbsp;{{../limit_estimate_enrollment}}
```
bexar: The first number is the actual count of students that would appear on class list - but i'm really not attached. Having the entire x of y clickable would increase link target size, which may be a good thing. I'm ok making that change.

klpittman: Cool. The larger click target is what I had in mind too.

PR #848, opened 2017-09-01
reset webservice_outage when reloading a page

fanglinfang:
https://jira.cac.washington.edu/browse/MUWM-3866

### DIFF
--- a/myuw/static/js/card/instructor_schedule/load_course_cards.js
+++ b/myuw/static/js/card/instructor_schedule/load_course_cards.js
@@ -3,18 +3,29 @@ var InstructorCourseCards = {
     dom_target: undefined,
     term: 'current',
 
-    render_init: function() {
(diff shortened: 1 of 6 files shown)


klpittman:
@fanglinfang is this related to a bug? How should I test this? Thanks!

klpittman:
I can't tell if this is a newly introduced bug, or if this was just always there even before you started this work. But it looks like a 404/500 from the GWS is causing MyUW to throw the 500 page instead of just showing the outage card. 

myuw_-_server_error_and_untitled

klpittman:
Just tested it on production and that is what happens there as well. For GWS, PWS, and SWS.

klpittman:
This appears to bring the original bug back. Testing on my-test with Chrome/mac. 
netid: thatssad

myuw__home

PR #855, opened 2017-09-07
Thatssad/muwm 3925

klpittman:
https://jira.cac.washington.edu/browse/MUWM-3925

Test with user: bill
Class: ESS 102 A spring 2013
Browser IE11

Also removes code where we were printing out two HTML elements with the same id="card_with_error" on the class list popup. Can view source for the class list popup and search for that ID to confirm there's only one now.


klpittman:
@fanglinfang ok, think I fixed it for real now. Take a look again, thx!

PR #860, opened 2017-09-09
Hotfix/muwm 3933

klpittman:
Not completely sure why this fixes the issue, but it seems to.  Also tightened up the print view after hearing some instructors complain that it wasted too much paper.


charlon:
Seems good.

PR #862, opened 2017-09-11
adding Autumn 2017 start date to VS.

wmwash:
can be viewed on branch (in Au quarter preview page) with
user: javerage
date: 2013-06-26 00:00:01


wmwash:
@klpittman agreed. it's a very quick fix. :-) I'll create the JIRA ticket you suggest - good call!

PR #863, opened 2017-09-11
Feature/transfer thrive

fanglinfang:
https://trello.com/c/6Tut9OnP


myuw/templates/handlebars/thrive.html
```diff
@@ -3,8 +3,20 @@
     <div class="container what-is-thrive">
         <h2 id="main" class="heading-page">What is Thrive?</h2>
         <div class="thrive-description">
+
+        {{#if target_fyp}}
             <p>Thrive is a series of weekly messages in MyUW designed to help you in your first year. While there are many things to do as you adjust to campus life, these just-in-time tips can help you know when and what to focus on during the coming months. Created by experienced students, faculty, and advisers, the messages provide pointers on how to explore opportunities and make the most of your learning experiences, meet people and connect to important campus resources, and create the foundation of your university career.</p>
 
+        {{else}}
+            <p>The weekly Thrive messages in MyUW are designed to help you adjust to campus life. Thrive messages for transfer students provide timely advice on what to focus on and when during your first quarter. Created by UW students, faculty, and advisers, the messages offer tips on how to
```
klpittman: @wmwash You need to close your <p> tag here, then have the <ul>. Below you will need to start a new <p> tag. Can't have lists inside of <p> elements.

myuw/templates/handlebars/thrive.html
```diff
@@ -3,8 +3,20 @@
...
+                <ul>
+                    <li>meet people and connect to important campus resources</li>
+                    <li>explore opportunities and make the most of learning experiences</li>
+                    <li>create a healthy foundation for your UW career</li>
```
klpittman: Also seems weird here that there is no final punctuation for this list, but the next line just starts a sentence. Not sure if you are looking for that type of editing/proof-reading.

fanglinfang:
@wmwash could you respond to Kevin's comment?

myuw/templates/handlebars/thrive.html
wmwash: LOL. I mentioned the same thing to Janice (her copy) but she didn't seem bothered. I'll fix it though.

myuw/templates/handlebars/thrive.html
wmwash: I was wondering about whether that was valid, but it rendered fine in Chrome. I should have double checked, my bad.

klpittman: No worries. Pretty much all browsers will render it properly. Technically it is bad HTML though. 

It is legal to leave off the </p> when followed by a <ul> (since a ul cannot be in the p tag, browsers essentially close the tag for you), so technically the error happens when there's no starting <p> tag following the <ul> and all of a sudden your browser sees the </p>.

PR #874, opened 2017-09-15
fixes MUWM-3968. only hides <h2>s of the heading class

klpittman:
Can test this on personal build by logging in as "bill" and viewing the Textbooks page. 

The heading "Teaching" should remain when going to mobile view, while the main page title above it becomes hidden.


fanglinfang:
The quarter and year (summer 2013) are gone, is that right?

PR #910, opened 2017-10-03
Hotfix/quicklinks reenable

devights:


myuw/test/views/not_test_link_redirect.py
```diff
@@ -1,5 +1,5 @@
 from myuw.test.api import missing_url, require_url, MyuwApiTest
```
fanglinfang: Please rename the file to test_link_redirect.py

PR #911, opened 2017-10-04
Hotfix/muwm 3997

fanglinfang:
How to verify, see https://jira.cac.washington.edu/browse/MUWM-3997

Kevin, the error message may not be appropriate. please check.


klpittman:
@fanglinfang I pushed some error message changes, but I was unable to test the "You do not have access" message. Which class is supposed to show that one?

Using "billsea, set date to 2017 autumn" I was only seeing the "Class not found" message.

PR #930, opened 2017-10-13
Bugfix/muwm 4012

klpittman:
https://jira.cac.washington.edu/browse/MUWM-4012

- Adjusts aria labels across a lot of pages
- Adds aria to Search link and area, switches it to Bootstrap for open/close feature.
- Adds aria to academic calendar filter buttons
- Fixes aria for Teaching page tabs


myuw/templates/calendar.html
```diff
@@ -9,7 +9,7 @@
 
 {% block content %}
 
-    <div id="main-content" role="main" aria-labelledby="main"></div>
+    <div id="main-content" role="main" aria-labelledby="main-content-label"></div>
```
fanglinfang: this main-content-label is on div whereas the Academics page is on h2?

klpittman: Yup, this basically means that the contents of this div is labelled by the h2 being referenced. So in this case, the main section of the page has a heading of "Academic Calendar"

PR #931, opened 2017-10-13
Removing additional ruleset for mobile redirection

ezturner:


myuw/views/page.py
```diff
@@ -55,23 +54,8 @@ def page(request,
...
-            log_exception(logger,
-                          '%s is_oldmyuw_mobile_user' % netid,
-                          traceback.format_exc())
-            logger.info("%s, redirected to legacy!" % netid)
-            return redirect_to_legacy_site()
-
-    else:
-        if is_oldmyuw_user():
-            return redirect_to_legacy_site()
+    if is_oldmyuw_user():
```
fanglinfang: You may want to remove def _is_mobile(request): ...

ezturner: I believe this is used in some other places such as : https://github.com/uw-it-aca/myuw/blob/master/myuw/logger/session_log.py#L27

fanglinfang: is_mobile = request.is_mobile or request.is_tablet  
this doesn't look like depending on the _is_mobile function defined in page.py

myuw/dao/__init__.py
```diff
@@ -21,9 +21,12 @@ def is_using_file_dao():
 
 
 def _is_optin_user(uwnetid):
-    file_path = _get_file_path("MYUW_OPTIN_SWITCH_PATH",
-                               "opt_in_list.txt")
-    return is_netid_in_list(uwnetid, file_path)
+    if is_using_file_dao():
```
fanglinfang: this is the better way!

PR #934, opened 2017-10-16
Feature/lti class list

jlaney:


myuw/views/lti/photo_list.py
```diff
@@ -0,0 +1,35 @@
+from django.utils.decorators import method_decorator
```
fanglinfang: could you add a test of this?

jlaney: Added.

PR #956, opened 2017-11-03
Bugfix/muwm 3138

klpittman:
Similar update to MUWM-3126, but for the financial aid notices on the tuition card on the Account page. Since this is an accordion interaction, the JS is slightly more complicated. Also updated the caret to use the new myuw-disclosurelink styles for consistency.


bexar:
I'm curious if we need some sort of separator, or spacing, or something else to differentiate the list of notices from the Financial Aid Status link. I thought at first the link was a list item then realized it was it's own thing. I don't think it's a blocker so I'll approve anyway, just something I noticed. 
screen shot 2017-11-03 at 1 23 27 pm

klpittman:
@bexar Completely agree. I also found the first link there a little confusing. I was just trying to keep to the scope of the bug here. Thanks for the review.

PR #957, opened 2017-11-06
Bugfix/muwm 2455

hlee2:
To fix the problem of the SR reads the empty final exam table before the visual schedule, the final exam content is moved after the visual schedule and h4 is applied.   
To address the limited browser/OS support of 'Aria-hidden=true', 'role=presentation' was added as an interim solution.

### DIFF
--- a/myuw/templates/handlebars/card/schedule/visual.html
+++ b/myuw/templates/handlebars/card/schedule/visual.html
@@ -4,155 +4,159 @@
 
 <div class="card"  data-name="VisualScheduleCard" data-type="card">
     {{#if total_sections}}
-    <!--Adding SR only final exam schedule-->
-    <h3 class="sr-only">Your {{ucfirst quarter}} {{year}} final exam</h3>
    <!-- {{# if has_early_fall_start }}
     <a href="#" class="show_full_term_meetings">Show {{ucfirst quarter}} Full Term Courses</a>
     <a href="#" class="show_efs_meetings" style="display: none;">Show {{ucfirst quarter}} Early Fall Start Courses</a>
     {{/if}}-->
-    <div class="final_schedule_table sr-only">
-	<table>
-	    <thead>
-	        <tr>
-	            <th scope="col">Class</th>
-	            <th scope="col">Exam date and time</th>
-	            <th scope="col">Exam location</th>
-	        </tr>
-	    </thead>
-	    <tbody>
-	        {{#each list_data}}
-	        <tr>
-	            <th scope="row">{{curriculum_abbr}} {{course_number}} {{course_title}}</th>
-                    {{#with final_exam}}
-<td>{{formatDateAsDate start_date}}, {{formatDateAsTimeAMPM start_date}} - {{formatDateAsTimeAMPM end_date}}</td>
-                    <td>
-	                {{#if building_tbd}}
-	                Room TBD
-	                {{else}}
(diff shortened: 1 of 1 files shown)


myuw/templates/handlebars/card/schedule/visual.html
```diff
@@ -4,155 +4,159 @@
...
-    {{/if}}
-    <!-- end - if there are meetings with times to be arranged -->
 
+    <h3 id="quarter-info" >
+            {{ucfirst quarter}} {{year}} Schedule
+            {{#if summer_term}}
+            <span class="shortTitle">{{summer_term}}</span>
+            {{/if}}
+    </h3>
+    <div><p class="sr-only"><a href="#CourseCards">Skip the visual schedule and go to course cards</a></p></div>
```
klpittman: This jump link location is no longer valid: `<a href="#CourseCards">` I think the correct location is: "/academics#academics_content_cards"

That being said, I'm not sure this is all that helpful though. The course cards are on a completely separate page which can be easily reached from the global nav.

myuw/templates/handlebars/card/schedule/visual.html
```diff
@@ -4,155 +4,159 @@
...
+    <div aria-hidden="true" role="presentation"><!-- Wrap the content under aria-hidden -->
+        <div id="schedule-tabs" class="schedule-period-module">
+    		<ul class="nav nav-pills schedule-period-list">
+            {{#each schedule_periods }}
+            <li class="{{#equal ../active_period_id @key}} active{{/equal }}"><a class="schedule-period-anchor" href="#" data-period_id="{{ @key }}"><span class="schedule-period">{{#equal "finals" @key}}Finals {{ else }}{{ toMonthDay start_date }} - {{ toMonthDay end_date }} {{/equal }}</span></a></li>
+            {{/each }}
+    		</ul>
+        </div>
+		<h4 class="sr-only">Your {{ucfirst quarter}} {{year}} class schedule</h4>
```
klpittman: It doesn't really make sense here to have a `<h4 class="sr-only">Your {{ucfirst quarter}} {{year}} class schedule</h4>` inside an `aria-hidden="true"` <div>.

You're hiding this entire content so you can remove all the "sr-only" stuff from it.

myuw/templates/handlebars/card/schedule/visual.html
```diff
@@ -4,155 +4,159 @@
...
+                </div>
+    	    </div>
+            {{/each}}
+        </div>
+        {{/if}}
+    </div>
+        <!-- end - if there are meetings with times to be arranged -->
+        
+        
+        <!--Adding SR only final exam schedule-->
```
klpittman: Since this is a temporary fix until Stephen is done with his work, why not just comment out this entire table? Seems weird to me to keep a screen-reader only table and content when the table is still empty. 

Might be better to hide the entire card from screen readers until it can get fixed properly. At this point the only content visible to screen readers is a few headings.

hlee2: yeah, I thought about that option too.

myuw/templates/handlebars/card/schedule/visual.html
hlee2: A good catch! 
Understood your point on the usefulness. Until we fixe the Visual schedule table, it can crutch users to the course schedule info,  IMO.

PR #958, opened 2017-11-09
Feature/remove banner

klpittman:
Should be checked across a handful of affiliations.


myuw/templates/base.html
```diff
@@ -152,54 +141,61 @@ <h2 class="sr-only" id="app_messages_desktop_label">MyUW Announcements</h2>
...
+                            {% endif %}
+                            <li><a {% if is_myuw_profile_page %}class="active"{% endif %} href="/profile/"><i class="icon-user icon-user-female" aria-hidden="true" style="font-size:14px;"></i>Profile</a></li>
+
+                        </ul>
 
+                    </div>
+                    {% endblock %}
+                    <!-- end app_navigation -->
+
+                    <!-- Unless applicant -->
```
klpittman: @ezturner Here is where we want to wrap an unless conditional around the welcome block.

myuw/templates/base.html
```diff
@@ -152,54 +141,61 @@ <h2 class="sr-only" id="app_messages_desktop_label">MyUW Announcements</h2>
...
 
-                <!-- app_navigation -->
-                {% block app_navigation %}
-                <div id="menu_container" class="myuw-menu" role="navigation">
+                <div class="myuw-nav-col">
 
-                    <ul>
+                    <!-- app_navigation -->
+                    {% block app_navigation %}
+                    <div id="menu_container" class="myuw-menu" role="navigation">
```
wmwash: is class myuw-menu still defined somewhere in a LESS file? to me it looked like that got changed to a different class name in bolilerplate.less..

myuw/static/css/typography.less
```diff
@@ -75,6 +75,19 @@ abbr[title] {
     font-size: @fontsize-small;
 }
 
+/** Can be used on an element or as a container for multiples **/
+.myuw-muted {
+    color: #777;
+
+    a&,
```
wmwash: I had to look this one up as I'd never seen this SASS/LESS syntax. But even now somewhat more informed, I'm not clear what you're trying to do. This compiles to `a.myuw-muted`  and `.myuw-muted a`, which both get the hover and visited style treatment you specified. But in the Welcome box only the later structure exists. The footer has `a.myuw-muted`, but not sure why you'd want to us that style in the footer where the links are white... Moreover, the Welcome links are blue by default...is that your intention?

myuw/templates/base.html
```diff
@@ -265,6 +261,7 @@ <h2 id="categories" class="sr-only">Categorized Resources</h2>
                         <li class="footer-links-list-item"><a href="//www.washington.edu/online/terms/">Terms</a></li>
                         <li class="footer-links-list-item"><a href="//www.washington.edu/online/privacy/">Privacy</a></li>
                         <li class="footer-links-list-item visible-xs-inline"><a href="/logout">Sign Out</a></li>
+                        <li class="footer-links-list-item visible-sm-inline"><a href="#" class="myuw-muted" data-toggle="modal" data-target="#optout_modal">Switch back to the old MyUW</a></li>
```
wmwash: I'm not seeing this render in the UI. I can see the code in the inspector (emulating a mobile device), but nothing visible in the UI. Is this an issue with my set up, or do you see the same?

myuw/static/css/typography.less
klpittman: What I was going for here is the ability to either use .myuw-muted as a class you could add to any text or link which would make it a muted gray. 

Additionally, I added the "&" stuff in there (which compiles to 'a.myuw-muted' and '.myuw-muted a') so that you could use the .myuw-muted as a wrapper class over many items. I figured that would be more efficient for when you want to have a collection of muted text/links so you don't have to add the class to each text/link.

I didn't intend to add the myuw-muted class to the link in the footer, so good catch there. 

I think links, in general, are blue by default. The myuw-muted class is intended to overwrite that to make them gray. Here's what I see on my build, which was my intent:
myuw__home

myuw/templates/base.html
klpittman: Yup, in the boilerplate there were two places it was defined. 

1. It was being used to set up the grid using 'display: flex' (flex boxes). line 50-60ish. I wrapped myuw-menu in another div called "myuw-nav-col" to create a left hand column which the nav and the welcome message could sit in. So on line 60, myuw-menu became myuw-nav-col as far as the grid is concerned.

2. myuw-menu was also defined lower in boilerplate.less. This portion does more of the menu styling and the responsive stuff for mobile vs. desktop. That hasn't changed and is still there. Line 341.

myuw/templates/base.html
klpittman: I'm not sure why I put one as visible-xs-inline and the other link as visible-sm-inline. I just changed them to both be visible-sm-inline, so both links should show up below about 1000px wide.
myuw__home

myuw/static/css/typography.less
wmwash: got it, that rationale makes sense to me.

myuw/templates/base.html
wmwash: got it. just didn't see that when i searched boilerplate.less. it may have been collapsed code.

PR #961, opened 2017-11-15
Hlee2/feature/applicant content

ezturner:


myuw/static/css/mobile.less
```diff
@@ -828,6 +820,52 @@ label { font-weight: normal; margin-bottom: auto; }
...
+		}
+	
+    // Application cards styles
+   .myuw-card.application-card { //remove top padding
+		    margin-top: 0;
+	    }
+
+    .application-card {
+
+	    .myuw-btn-big {
```
klpittman: `myuw-btn-big`

If we plan to use this style of button elsewhere, we should probably move it to the components.less file and not nest it here under .application-card. As written, these styles only apply for the application-card, but its name suggests it is a widely used style.

klpittman: I'm also a little confused how it relates to `card-action-btn`. You're using them together which suggests the `myuw-btn-big` is sort of an extension of `card-action-btn`, but the naming scheme is different. It would maybe make more sense to use "myuw-btn" and "myuw-btn-big". I would put both in the components.less.

If the `myuw-btn-big` is just a larger `card-action-btn`, I'm wondering why the switch to Roboto typeface?

myuw/templates/handlebars/card/seattle_application.html
```diff
@@ -0,0 +1,116 @@
...
+                <div class="col-sm-3">
+                    <div class="myuw-card-row-heading">Mail</div>
+                </div>
+                <div class="col-sm-9">
+                    <div class="myuw-card-row-content">Registration and Transcripts Office<br>University of Washington <br>Box 355850 <br>Seattle, WA 98195-5850</div>
+                </div>
+            </div>
+	</div>
+	<div class="myuw-card-section" style="padding-top: 8px;">
+		<div>
```
klpittman: Not sure, but it doesn't seem like this <div> is used for anything.

myuw/templates/handlebars/card/welcome_card.html
```diff
@@ -0,0 +1,18 @@
...
+{% tplhandlebars "welcome_card" %}
+<div class="myuw-card">
+    <h3 class="myuw-card-title">
+        Welcome to MyUW, a student portal
+    </h3>
+
+    <div class="myuw-card-section last">
+        <p>MyUW is your go-to place for application status, UW resources for applicants, your account information, and much more.</p>
+        <h4 class="myuw-card-heading">In MyUW you can:</h4>
+	        <ul style="padding-left:20px">
```
klpittman: Not critical for this release, but we should probably standardize how we want bulleted lists to look and add that to the style. Adjusting the left padding on just this one seems strange to me.

myuw/templates/handlebars/card/bothell_application.html
```diff
@@ -0,0 +1,66 @@
...
+
+{{else}}
+    <h3 class="myuw-card-title">
+        Your Bothell Application for {{toTitleCase quarter}} {{year}}
+    </h3>
+
+    <div class="myuw-card-section">
+
+        <div class="clearfix">
+             <span class="btn btn-default card-action-btn myuw-btn-big"><a href="https://admissions.uwb.edu/apply/status"  data-linklabel="Your application status details">View your first year application status and checklist</a></span>
```
klpittman: For reference, here is where I'm seeing the two btn classes used together.

`<span class="btn btn-default card-action-btn myuw-btn-big">`

myuw/templates/handlebars/card/seattle_application.html
hlee2: it overwrites the padding to look more consistent between chunks.

myuw/templates/handlebars/card/welcome_card.html
hlee2: I couldn't find any other place with a bullet list, so I threw the first one here.  The bootstrap's left padding looked too big to me.

myuw/templates/handlebars/card/seattle_application.html
klpittman: Sorry, github ate my formatting. I was referring to that `<div>`.

myuw/static/css/mobile.less
hlee2: Your are right that .myuw-btn-big is an extension of .card-action-btn. 
The Roboto typeface looks nicer in the button - more refined button for applicant! 
.myuw-btn-big is moved to component.less.

Correcting the .card-action-btn will require more templates changes and not critical for this release, so will add to the backlog.

klpittman: IMO, if we like Roboto for buttons we should just use it for all the buttons, not just the big ones.

When you tackle refactoring the buttons, I'd recommend only building on the `btn` class instead of both `btn` and `btn-default`. You aren't using a lot of the `btn-defalut` styles and I think it'd be better to just make the `myuw-btn` more complete.

"btn btn-default card-action-btn myuw-btn-big

PR #978, opened 2017-11-27
Attempting to fix possible race condition

ezturner:


ezturner:
see: https://github.com/uw-it-aca/myuw/pull/973/files#diff-eb0dfddd0fc87db6a86e591eb835d7e4R1

PR #1004, opened 2017-12-21
resolves MUWM-4077

klpittman:
Changes implementation for hiding footer links at desktop size. Creates a myuw specific class similar to bootstraps hide/show helper utility classes.


charlon:
Well that was easy!

PR #1007, opened 2017-12-26
Feature/thrive next server

devights:


myuw/templates/handlebars/husky_experience.html
```diff
@@ -0,0 +1,181 @@
...
+                        </div>
+
+                    </div>
+
+
+
+                </div>
+            </div>
+        </div>
+        -->
```
ezturner: Why is this commented out?

myuw/dao/hx_toolkit_dao.py
```diff
@@ -0,0 +1,93 @@
...
+    quarter = term.quarter
+
+    if quarter == 'autumn':
+        year += 1
+
+    phases = ['A', 'B', 'C']
+    start_year = 2009
+    years_since_start = year - start_year
+    phase_offset = 0
+    if years_since_start > 0:
```
ezturner: Is this if statment necessary? Won't 0% 3 be 0?

PR #1038, opened 2018-01-24
Fang/gws

fanglinfang:
The delta in speed test drops from ~4 sec to 2.5 sec.
This is built on my-test.


myuw/dao/affiliation.py
```diff
@@ -303,8 +195,6 @@ def get_identity_log_str(request):
         res += ' Bothell'
     if affi["tacoma"] or affi["official_tacoma"]:
         res += ' Tacoma'
-    if affi["official_pce"]:
```
fanglinfang: there isn't a way to determine official_pce

PR #1097, opened 2018-02-15
Bug/major refactor

ezturner:


fanglinfang:
Building this branch on my-test

myuw/dao/student.py
```diff
@@ -0,0 +1,129 @@
+"""
+This module will provide a
+"""
+from uw_sws.enrollment import enrollment_search_by_regid
```
fanglinfang: Please change to use myuw.dao.enrollment: enrollment_search, which is prefetched in the request

myuw/dao/student.py
```diff
@@ -0,0 +1,129 @@
...
+    attribute, and a 'current' and 'rollup' major field
+    :param regid
+    :return: dict
+    """
+    enrollments = enrollment_search_by_regid(regid)
+
+    return _get_majors(enrollments)
+
+
+def get_rollup_and_future_majors(majors):
```
fanglinfang: Would it be OK to change "rollup" to "last" (to be more specific)?

ezturner: The reason rollup is called rollup is because we default to current, then go to future, then go to last

myuw/dao/student.py
```diff
@@ -4,7 +4,7 @@
 determinations about user context in the case that a user is not
 enrolled for the current quarter
 """
-from uw_sws.enrollment import enrollment_search_by_regid
+from myuw.dao.enrollment import enrollment_search_by_regid
```
fanglinfang: please change enrollment_search_by_regid to "enrollment_search"

ezturner: Ah I see, we're caching withing the request. I'll update the method signatures to take a request instead and call enrollment_search.

myuw/dao/student.py
```diff
@@ -0,0 +1,132 @@
...
+def get_student_status(request):
+    """
+    This will return a dictionary with the student status over time alongside
+    various rollup variables
+    :param request:
+    :return: dict
+    """
+    enrollments = enrollment_search(request)
+
+    return {
```
fanglinfang: Sorry for being so high maintenance. 
Could you refactor the code to go through the enrollments once, and gather the major, minor, at the same time?

ezturner: Yeah, that's a good idea! Thanks for the tip.

myuw/dao/student.py
```diff
@@ -0,0 +1,143 @@
...
+This module will provide a method to retrieve all user majors, class standings
+and minors along with current and rollup variables to facilitate
+determinations about user context in the case that a user is not
+enrolled for the current quarter
+"""
+from myuw.dao.enrollment import enrollment_search
+from uw_sws.term import get_current_term
+
+
+def get_student_status(request):
```
fanglinfang: Could you tell me why we need this function? It seems only being referenced in test.dao.

myuw/dao/student.py
```diff
@@ -0,0 +1,143 @@
...
+
+def get_rollup_and_future_minors(minors):
+    return _get_rollup_and_future(minors['minors'], minors['rollup'])
+
+
+def _get_minors(enrollments):
+    return _process_fields(enrollments, ["minors"])['minors']
+
+
+def get_class_standings(request):
```
fanglinfang: Could you tell me where this function is used?

myuw/dao/student.py
ezturner: I'm planning on using get_student_status to refactor the profile card and provide this information on the clientside.

myuw/dao/student.py
ezturner: I don't have a current use for get_class_standing but I had thought that there may be a future feature wherein we run into a similar issue where class standing is not available to a module due to a lack of current enrollment data

myuw/dao/student.py
```diff
@@ -0,0 +1,129 @@
...
+        status[attribute][attribute] = {term: getattr(enrollments[term],
+                                                   attribute)
+                                     for term in enrollments}
+
+    if 'current' not in status[attributes[0]]:
+        sorted_terms = sorted(enrollments.keys())
+
+        has_future_entry = False
+
+        for term in sorted_terms:
```
fanglinfang: going through each term

myuw/dao/student.py
```diff
@@ -0,0 +1,129 @@
...
+                for attribute in attributes:
+                    status[attribute]['rollup'] = getattr(enrollments[term],
+                                                       attribute)
+                has_future_entry = True
+                break
+
+        recent_term = sorted_terms[0]
+
+        if not has_future_entry:
+            for term in sorted_terms:
```
fanglinfang: going through each term another time

myuw/dao/calendar_mapping.py
```diff
@@ -12,40 +15,40 @@
...
-    return _get_calendars(request, _get_enrollments(request))
+
+    try:
+
+        majors = get_majors(request)['rollup']
+        majors = [major.major_name for major in majors]
+
+        minors = get_minors(request)['rollup']
+        minors = [minor.short_name for minor in minors]
+
```
fanglinfang: Can we get {
        'majors': majors,
        'minors': minors
    }
in one call?

myuw/dao/calendar_mapping.py
```diff
@@ -12,40 +15,28 @@
 
 
 def get_calendars_for_current_user(request):
-    return _get_calendars(request, _get_enrollments(request))
+
+    enrollments = {
+        'majors': [maj.major_name for maj in get_majors(request)['rollup']],
+        'minors': [mino.short_name for mino in get_minors(request)['rollup']]
+    }
```
fanglinfang: The original code only loop through the enrollments once to get both major and minor

PR #1119, opened 2018-02-24
Bug/muwm 3954

ezturner:
Adding an alert to tell users running IE in document mode 7 why MyUW will not work and to switch to a different browser.


myuw/templates/base.html
```diff
@@ -59,6 +59,15 @@
         <!-- android icons -->
         <link rel="shortcut icon" sizes="196x196" href="{% static 'images/android-196x196.png' %}">
 
+        <script>
+            // Check if this browser is Internet Explorer 7 and running Document Mode 7
+            if(document.documentMode !== null && document.documentMode === 7){
```
devights: How about documentMode >= 9, in my testing that's the first version that at least loads the card content (but leftnav is screwed up)?

ezturner: Sure

PR #1120, opened 2018-02-24
Bug/muwm 3954

ezturner:


myuw/templates/base.html
```diff
@@ -59,6 +59,15 @@
         <!-- android icons -->
         <link rel="shortcut icon" sizes="196x196" href="{% static 'images/android-196x196.png' %}">
 
+        <script>
+            // Check if this browser is Internet Explorer and running Document Mode 9 or lower
+            if(document.documentMode !== null && document.documentMode <= 9){
+                // Notify the user that we do not support this document mode
+                alert("We have detected that your browser is running in document mode for Internet Explorer 9 or lower.\n" +
+                      "Please switch to Firefox or Chrome.")
```
klpittman: Wonder if we should say something like: 

"We have detected that your browser is running in document mode for Internet Explorer 9 or lower.

Please set document mode to 10 or higher, or switch to another browser."

klpittman: Mainly I want folks to know there is an option that will allow them to continue using IE, just in case that's the only browser they have available to them.

PR #1126, opened 2018-03-05
Fixing MUWM-4144

ezturner:


ezturner:
Won't these lines accomplish the same result?
https://github.com/uw-it-aca/myuw/blob/e3f2b0cdd7cb370cc29855880f23146905dc9266/myuw/static/js/card/bothell_application.js#L6-L12

fanglinfang:
Ah, yes!

PR #1130, opened 2018-03-06
Feature/c2 tuitioncard

devights:


myuw/templates/handlebars/card/accounts/tuition.html
```diff
@@ -3,115 +3,133 @@
...
+		            <span class="card-badge-action pull-right">in {{tuition_date_offset}} days</span>
+		        </div>
+		    </li>
+    {{/if}}<!-- end tuition_date -->
 	</ul><!-- .end of the balance list -->
 
+    {{#unless is_c2_grad}} <!-- Access msg for non-PCE -->
     <div class="card-related-messages">
         <p><a href="https://sdb.admin.uw.edu/sisStudents/uwnetid/release.aspx" target="_blank">Give access to your tuition account and financial aid information</a> to parents or other third parties.</p>
+	{{/unless}}
```
klpittman: Double check the order of your closing statements here, it looks like you may have a bug. 

Above you are opening the `{{unless is_c2_grad}}`, then you open a `<div>` and `<p>` tag.  The `<p>` gets closed but the `<div>` does not before the `{{/unless}}`. This will likely cause you to have an extra `</div> in your code when the user "is_c2_grad".

klpittman: I'm also a little confused about who should see it because the comments says "non-PCE", but then the variable name is "is_c2_grad" which implies undergrad PCE students should see it. I didn't see this message when logging in as any of the undergrad PCE users.

myuw/templates/handlebars/card/accounts/tuition.html
```diff
@@ -3,115 +3,133 @@
...
+        {{else}}  <!-- else, there is a balance -->
 
-	            {{/if}}
-                <div class="card-badge clearfix middle-row">
-                    <div class="pull-right" >
-                        <div class="btn btn-default card-action-btn"><a href="http://f2.washington.edu/fm/sfs/tuition/payment" target="_blank" data-linklabel="Make Tuition Payment">Make payment</a></div>
+	    	<li>
+                <div class="card-badge clearfix" style="padding-bottom:0">
+                    <div class="pull-left">
+                        <h4 class="card-badge-label">Amount Due <br><span style="font-size:14px;font-weight:normal;line-height:32px;">Student Fiscal Services</span></h4>
```
klpittman: `<span style="font-size:14px;font-weight:normal;line-height:32px;">`

This inline css appears a few times in this file and should probably be refactored into a CSS class.

myuw/templates/handlebars/card/accounts/tuition.html
hlee2: {{unless is_c2_grad}} moved after <div>. 
jpce , C2 undergrad displayed the message. jinter, C2 grad didn't on my build. Will double check on my-test going forward.

PR #1132, opened 2018-03-07
resolves MUWM-3812

devights:


devights:
Adding additional logging to for base schedule exceptions

PR #1139, opened 2018-03-14
Feature/teaching page

fanglinfang:
Test spec: https://docs.google.com/document/d/14-p4tLjrUMjX7kTGUXFQ8bcw4M502aZamUmSvWRw9uo/edit
+ course color bug fix
+ Class list opens in a new tab instead of in a popup.

New code test coverage:
 dao/user_course_display.py 100%
 api/instructor_section_display.py 100%

The JS side tests will be done separately.


myuw/views/api/instructor_section_display.py
```diff
@@ -0,0 +1,43 @@
...
+from myuw.views.error import handle_exception
+
+logger = logging.getLogger(__name__)
+
+
+class CloseMinicard(ProtectedAPI):
+
+    def get(self, request, *args, **kwargs):
+        """
+        GET returns 200 with email lists for the course
```
devights: Looks like these comments aren't valid

PR #1153, opened 2018-03-21
Feature/bookstore api update

ezturner:


myuw/templates/handlebars/textbooks.html
```diff
@@ -189,7 +189,7 @@ <h4 class="book-title">{{title}}</h4>
 
 
 
-    <div style="text-align:center; margin-bottom:10px;"><a target=_blank href="http://www.ubookstore.com/adoption-search"><button class="btn btn-default card-action-btn" type="button">Start textbook shopping</button></a></div>
+    <div style="text-align:center; margin-bottom:10px;"><a target=_blank href="{{#if order_url }} {{ order_url }} {{else}}http://www.ubookstore.com/adoption-search{{/if}}"><button class="btn btn-default card-action-btn" type="button">Start textbook shopping</button></a></div>
```
klpittman: having a button tag inside an `<a>` is kind of bad form. I propose just using the link:

`<a target="_blank" class="btn btn-default card-action-btn" href=" http://www.ubookstore.com/adoption-search-results?ccid=22234,23547,23757 ">Start textbook shopping</a>`

I'm also noticing there are just extra space characters before and after the URL in the `href`. Probably won't cause any issues but seems easy to clean up.

PR #1162, opened 2018-03-23
refresh the page if api call succeeded.

fanglinfang:


myuw/static/js/card/instructor_schedule/load_course_cards.js
```diff
@@ -83,6 +83,7 @@ var InstructorCourseCards = {
     },
 
     _render: function () {
+        console.log("InstructorCourseCards _render");
```
ezturner: Do we still need this log statement?

fanglinfang: good catch!

PR #1179, opened 2018-04-05
Fang/onboarding

fanglinfang:
1. replace individual email icon and title by a single one
2. banner display and close
3. all users get new MyUW by default. Only those opt-out will be route to legacy
4. partially handle uwnetid update


myuw/models/__init__.py
```diff
@@ -340,3 +327,39 @@ class Meta(object):
             ["user", "section_label"],
         ]
         ordering = ['section_label']
+
+
+class MigrationPreference(models.Model):
+    user = models.OneToOneField('User', on_delete=models.CASCADE)
```
ezturner: Would it be good to add an index on this field?

fanglinfang: models.OneToOneField implies unique=True, this means Django will automatically create an index for this field

ezturner: Interesting, thank you!

PR #1186, opened 2018-04-12
Feature/staff content

devights:


myuw/templates/base.html
```diff
@@ -366,7 +335,7 @@ <h2 id="categories" class="sr-only">Categorized Resources</h2>
         <script src="{% static 'js/cards.js' %}"></script>
         <script src="{% static 'js/future.js' %}"></script>
         <script src="{% static 'js/navbar.js' %}"></script>
-        <script src="{% static 'js/thrive.js' %}"></script>
```
fanglinfang: We don't want to remove thrive, don't we?

myuw/models/__init__.py
```diff
@@ -368,3 +368,19 @@ def __str__(self):
...
         db_table = 'migration_preference'
+
+
+class ResourceCategoryPin(models.Model):
+    resource_category_id = models.CharField(max_length=255)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+
+    @staticmethod
+    def get_user_pinned_categories(user):
+        try:
```
fanglinfang: To increase test coverage, you may want to change to:
if ResourceCategoryPin.objects.filter(user=user).exists():
   ...

myuw/migrations/0009_resourcecategorypin.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
```
fanglinfang: could you merge the three migration files into one?
just remove them and run makemigrations once.

PR #1213, opened 2018-04-18
Bugfix/muwm 4212

hlee2:
Dawg dollar + Manage account button is added to Husky card for Tacoma campus affiliated people.


myuw/templates/handlebars/card/accounts/husky.html
```diff
@@ -7,8 +7,24 @@ <h3 aria-hidden="true">Husky Card</h3>
...
+                <div class="pull-left">
+                    <h4 class="card-badge-label" aria-hidden="true">Tacoma Dawg Dollars</h4>
+                    <h4 class="sr-only">Tacoma Dawg Dollars Account</h4>
+                </div>
+                <div class="pull-right">
+                    <div class="btn btn-default card-action-btn" ><a href="http://apps.tacoma.uw.edu/dawgdollars/" target="_blank" data-linklabel="Tacoma Dawg Doallar account">Manage account</a></div>
+                </div>
+            </div>
+        </li>
+     </ul>
```
klpittman: Not a release stopper, necessarily, but I think this card should just have a single `<ul>`. It is a little strange to have a "list" with a single item in it (tacoma dawg dollars), and then to follow it with another, separate list with the Husky Card balance information in it. It looks like the list items are used elsewhere to enumerate whatever is labelled in the card. 

A single list would be more semantic and better for accessibility.

myuw/templates/handlebars/card/accounts/husky.html
```diff
@@ -45,15 +60,61 @@ <h4 class="sr-only">Employee Husky Account</h4>
...
-                <div class="btn btn-default card-action-btn pull-right"><a href="{{employee_husky_card.add_funds_url}}" target="_blank">Add funds to your accounts</a></div>
-        {{/if}}
-
+            {{else if employee_husky_card}}
+                <div class="btn btn-default card-action-btn pull-right" style="margin-top:0px;"><a href="{{employee_husky_card.add_funds_url}}" target="_blank" aria-label="Manage Husky account">Manage account</a></div>
+            {{/if}}
+            </li>
+        </ul>
+        <!-- Non-Tacoma Husky card case  -->
+        {{else}}
```
klpittman: For future reference, you can probably refactor this template down so that you don't have to repeat this whole bottom section for the non-tacoma case. 

It looks like the only difference is where the last Manage Account button is placed, so you could put `{{#if is_tacoma}}` conditionals around the part that needs to change at the bottom. Basically if tacoma, add `<li>` around the button, otherwise close the `</ul>` and add the button below.

myuw/templates/handlebars/card/accounts/husky.html
hlee2: Changed it to single <ui> for the Dawg Dollars account and Husky accounts.

myuw/templates/handlebars/card/accounts/husky.html
hlee2: Refactored.

PR #1214, opened 2018-04-18
resolves VS finals display

devights:


devights:
This helper got removed during a refactor.

PR #1221, opened 2018-04-19
wraps offending call in try

devights:


devights:
Will wait to merge into master until GO/NOGO meeting

PR #1224, opened 2018-04-20
Finalizing WSData cache clear and test

ezturner:


fanglinfang:
merged this into fang/js-coverage, let's see if it fix the test

PR #1228, opened 2018-04-26
adds required GWS v2 resources

jlaney:


fanglinfang:
MyUW doesn't need these. We have removed effective_member/ mock data.

PR #1230, opened 2018-04-26
styles global link hover, focus, and active states. resolves MUWM-2105

klpittman:
Scope is limited to the default link styles. Does not include various other links in the nav, boilerplate, styles of links that are not the standard.


myuw/static/css/typography.less
```diff
@@ -33,7 +33,15 @@ hr {
     border-top: 1px solid #ddd;
 }
 
-/** Lists **/
+/*** Links ***/
+a { text-decoration: none; color: #4d66ca;
+    &:hover, &:focus { text-decoration: underline; color: #334baa; }
+    &:active { text-decoration: underline; color: @uw-metalic-gold; }
```
wmwash: All looks good, save for this gold color. Better to use a blue similar to #2e55f2 - a bright blue.

PR #1240, opened 2018-05-04
Develop

ezturner:


myuw/static/css/boilerplate.less
```diff
@@ -455,3 +455,15 @@ body { padding: 0; margin: 0; font-family: "Open Sans",Helvetica,Arial,Verdana,s
...
+
+/*** Margin Override ***/
+
+
+.marginless {
+    margin-bottom: 0 !important;
+    margin-top: 0 !important;
+    margin-left: 0 !important;
+    margin-right: 0 !important;
+    margin: 0 !important;
```
klpittman: All you really need is the ` margin: 0 !important;` line. It's a shorthand for all of the above lines.

PR #1252, opened 2018-05-09
Thatssad/notice enhancement

klpittman:
* pulls notice LESS into its own file. 
* Removes lots of unused CSS
* handlebars helper for printing the wrapping <ul> if there are critical notices OR regular notices.
* combines list of notices into single list, critical on top.
* removes icons

### DIFF
--- a/myuw/templates/handlebars/notices.html
+++ b/myuw/templates/handlebars/notices.html
@@ -175,7 +175,7 @@
                 {{ /if }}
                 <!-- end collapsy 3 -->
 
-                {{ #if next_week.notices }}
+                {{ #if next_week.tices }}
                 <!-- collapsy 4: next week -->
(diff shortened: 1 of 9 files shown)


myuw/templates/handlebars/notices.html
```diff
@@ -175,7 +175,7 @@ <h2 id="main" class="heading-page">Notices</h2>
                 {{ /if }}
                 <!-- end collapsy 3 -->
 
-                {{ #if next_week.notices }}
+                {{ #if next_week.tices }}
```
devights: typo?

klpittman: good catch. I didn't realize i even edited this page. Might have been a find and replace issue. Thanks!

PR #1269, opened 2018-05-15
Feature/docker

ezturner:


ezturner:
This is my first shot at Dockerizing our build process - I didn't end up getting a trivial setup with maximum parity and ease of use working quite as well as I had hoped - there's no settings sync and getting your DB backed up is going to be a bit of a pain for now - but we'll have one click builds for designers relatively easily and I can iterate on it over time. I'll also probably pick up some neat tricks at Dockercon.

ezturner:
Instructions: https://github.com/uw-it-aca/myuw/wiki/Installation-with-Docker

PR #1280, opened 2018-05-25
Removing usage of parse_date to use Moment parsing

ezturner:


ezturner:
This is built on my-dev

PR #1322, opened 2018-06-14
Resolve MUWM-4292

fanglinfang:
https://jira.cac.washington.edu/browse/MUWM-4292

Couldn't find a way to unit-test the IntegrityError. Suggestion will be appreciated.

### DIFF
--- a/myuw/dao/user_course_display.py
+++ b/myuw/dao/user_course_display.py
@@ -103,7 +103,7 @@ def _save_section_color(user, section, color_id):
             logger.error("%s (%s %s color_id: %d)", ex, user,
                          section_label=section_label,
                          color_id=color_idex)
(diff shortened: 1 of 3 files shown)


myuw/dao/user_course_display.py
```diff
@@ -91,14 +92,19 @@ def _save_section_color(user, section, color_id):
     Store the color of the section in DB
     """
     section_label = section.section_label()
-    if UserCourseDisplay.exists_section_display(user, section_label):
-        _update_color(user, section_label, color_id)
```
fanglinfang: I eliminated the _update_color call at line#95 as the function _save_section_color is only called when the section_label is not in the existing_color_dict. Is it OK if the record is created by another process at almost the same time. For the same schedule the colors are deterministic.

PR #1336, opened 2018-06-20
Fix/summer sub term

ezturner:


ezturner:
This is built on my-dev

myuw/static/css/cards.less
```diff
@@ -268,8 +268,15 @@
         a:hover { background-color: #ddd; }
     }
 
-} /*** END myuw-cards ***/
+    .myuw-card-subheader {
```
klpittman: @DannyNineNineNine not sure we really need a new style here for this. Could we not just use the `myuw-card-heading` existing class?

myuw/templates/handlebars/card/summary/schedule.html
```diff
@@ -2,7 +2,7 @@
     {% tplhandlebars "instructor_summary_schedule" %}
     <div class="myuw-card teaching-schedule-summary" data-type="card"
          data-name="SummaryScheduleCard" data-identifier="{{ title }}">
-         <div style="position:relative;">
+         <h4 style="position:relative;">
```
klpittman: @DannyNineNineNine why did this get changed to an `h4`?

myuw/templates/handlebars/card/summary/schedule.html
```diff
@@ -21,56 +21,44 @@ <h3 class="myuw-card-title">
...
-                                                   {{ curriculum_abbr }} {{course_number}} {{section_id}}</a>
-                                            </div>
 
-                                            <div id="secondary_{{section_label}}" class="collapse secondary-section-disclosure"
-                                             rel="{{ curriculum_abbr }} {{course_number}} {{section_id}}"
-                                             aria-label="Secondary sections for {{ curriculum_abbr }} {{course_number}} {{section_id}}">
+             {{#if is_summer}}
+
+                {{#if full_term }}
+                 <div class="myuw-card-subheader">
```
klpittman: I think this should be `<h4 class="myuw-card-heading">` instead of a div. It is acting as a header so semantically that's the right tag. We can specifically adjust the top margin if it is too big.

myuw/static/css/cards.less
DannyNineNineNine: @klpittman: that was done by @ezturner. There is minimal visual differences (size, padding), so unless there's a reason Ethan can explain, I have no objections to switching the class and having a new one.

klpittman: Ah ok. Let me know if you need help making these changes.

myuw/templates/handlebars/card/summary/schedule.html
ezturner: That was me - an accident, maybe my IDE?

myuw/templates/handlebars/card/summary/schedule.html
ezturner: Great, I wasn't sure if using a div or header would be better. I was also curious if we needed to use the class - aren't all headers within a card going to look like this or have a separate class?

PR #1338, opened 2018-06-21
Remove the quarter name capitalization style

hlee2:
Capitalize the quarter name in Textbooks page


myuw/templates/handlebars/textbooks.html
```diff
@@ -2,7 +2,7 @@
     {% tplhandlebars "textbooks" %}
 
 <div class="container textbooks-container">
-    <h2 id="future_cards_header" class="myuw-heading-page"><span style="text-transform:uppercase;">{{quarter}} {{year}}</span> {{summer_term}} Textbooks</h2>
+    <h2 id="future_cards_header" class="myuw-heading-page"><span style="text-transform: capitalize;">{{quarter}} {{year}}</span> {{summer_term}} Textbooks</h2>
```
klpittman: Really minor, but I think we should use the Bootstrap class here instead of an inline style. It's a little cleaner.

`<span class="text-capitalize">`

klpittman: Other than that, fix looks good!

hlee2: Good catch, Kevin! Thx!

hlee2: Changed to bootstrap class

PR #1345, opened 2018-06-22
Modifying if statement for summer quarter split

ezturner:


ezturner:
This is built on my-dev

PR #1364, opened 2018-07-02
Bootstrap disclosures

klpittman:
Refactor of `toggle_card_disclosure` in myuw_m.js and subsequent cards/JS that were using it. Changes it to rely on Bootstraps disclosure code to handle the hide/show/animations and some of the attributes that change (aria-expanded, aria-hidden). Also changed it to use the same <a> tag for both hide and show, instead of using two separate <a> tags and wrapping elements as it was before. `data-` attributes are used to control the aria-labels, text, and titles to be used during hide/show.

* Events card on homepage > More events disclosure
* Grade card on Academics
* 3 disclosures on the reg card
 * Holds
 * MyPlan courses
 * Not Ready courses
* Evaluations on the Academics page


klpittman:
@ezturner Would love your review on this when you have some time, as I'm not super proficient with JS. Let me know if anything looks weird in here and if you have any suggestions. 

Note: I still need to test this across different browsers, especially IE. Will do that tomorrow. Thanks.

PR #1365, opened 2018-07-02
Feature/saml2

ezturner:


docker/settings.py
```diff
@@ -171,3 +171,14 @@
 SUPPORTTOOLS_PARENT_APP_URL = "/"
 
 REMOTE_USER_FORMAT = "uwnetid"
+
+MOCK_SAML_ATTRIBUTES = {
+    'uwnetid': ['javerage'],
+    'affiliations': ['student', 'member'],
+    'eppn': ['javerage@washington.edu'],
+    'scopedAffiliations': ['student@washington.edu', 'member@washington.edu'],
+    'isMemberOf': ['u_test_group', 'u_test_another_group'],
```
fanglinfang: It'd be better that this 'affiliations' mock data matches the mock PWS data
"EduPersonAffiliations":["member","student","alum","staff","employee"],

'isMemberOf':  ['u_astratst_myuw_test-support-admin']

travis-ci/settings.py
```diff
@@ -138,3 +138,14 @@
 COMPRESS_ROOT = "compress_root"
 RESTCLIENTS_TEST_MEMCACHED = True
 RESTCLIENTS_MEMCACHED_SERVERS = ('localhost:11211', )
+
+MOCK_SAML_ATTRIBUTES = {
+    'uwnetid': ['javerage'],
+    'affiliations': ['student', 'member'],
```
fanglinfang: Same as above

fanglinfang:
How about bill, which is another user having override permission?

Are 'affiliations' and 'scopedAffiliations' come by default? 
As they will be sticky to the original user, is it possible to turn them off?

ezturner:
It would actually appear that as long as you're using the Django development server you can still set REMOTE_USER just fine.

ezturner:
We would encounter some issues if we were loading in data from the SAML assertions (say group membership) but we are not at the moment.

PR #1372, opened 2018-07-03
Feature/travis cache

ezturner:
Currently our travis setup does essentially the following:

1. Install python dependencies and run migrations
2. Install node.js dependencies
3. Run JS tests.
4. Lint JS
5. Run pycodestyle
6. Run django tests.

With this setup, we can achieve the following:

Python Container:
1. Install new python dependencies, refresh old ones from cache
2. Run migrations
3. Run pycodestyle
4. Run tests

Node.js container:
1. Install new node.js dependencies, refresh old ones from cache
2. Lint JS
3. Run Tests:

Coveralls container (after):
1. Combine coverage files from the two runs
2. Upload to coveralls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 sudo: false
 language: python
 cache:
-  - pip
-  - bundler
+  directories:
+    - ~/.cache/pip
+    - node_modules
 services:
   - memcached
   - mysql
@@ -11,8 +12,8 @@ python:
   # - "3.6"
 env:
   matrix:
-    - DB=sqlite3
-    # - DB=mysql
+    - TEST=nodejs
+    - TEST=python
 before_script:
   - if [ "$DB" = "mysql" ] ; then mysql -u root -e 'create database myuw'; fi
   - if [ "$DB" = "mysql" ] ; then mysql -u root -e 'create user "myuw" identified by "my_pass"'; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,44 +1,42 @@
 sudo: false
-language: python
-cache:
-  directories:
-    - ~/.cache/pip
-    - node_modules
-services:
-  - memcached
-python:
-  - "2.7"
-  # - "3.6"
-env:
-  matrix:
-    - TEST=nodejs
-    - TEST=python
-before_script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,43 +1,69 @@
...
-language: python
-services:
-  - memcached
-  - mysql
-python:
-  - "2.7"
-  # - "3.6"
-env:
-  matrix:
-    - DB=sqlite3
```
fanglinfang: we no longer use sqlite3

.travis.yml
```diff
@@ -1,43 +1,69 @@
...
+    cache: pip
+    services:
+      - memcached
+    python:
+      - 2.7
+#     - "3.6"
+    before_script:
+#     - if [ "$DB" = "mysql" ] ; then mysql -u root -e 'create database myuw'; fi
+#     - if [ "$DB" = "mysql" ] ; then mysql -u root -e 'create user "myuw" identified by "my_pass"'; fi
+#     - if [ "$DB" = "mysql" ] ; then mysql -u root -e 'grant all on *.* to myuw'; fi
```
fanglinfang: no more DB set up

.travis.yml
```diff
@@ -1,43 +1,69 @@
...
-  - python manage.py migrate
-  - npm list
 
-script:
-  - pycodestyle myuw/ --exclude=migrations,myuw/static
-  # Find </br> tags
-  - grep -re '<\s*/\s*br\s*>' myuw/templates/ ; test $? -eq 1
-  - grep -r 'static/' myuw/ | grep -v /test/ | grep -v washington.edu/static; test $? -eq 1
-  - npm run jshint myuw/static/js/  --verbose
-  - npm run mocha myuw/static/js/test/ --recursive
```
fanglinfang: no longer run mocha?

ezturner: Mocha still gets run, but in it's own container optimized for node.js so that we can parallelize the test runs.

.travis.yml
ezturner: I've removed the sqlite3/mysql config option since we're no longer using mysql - sqlite is now the default.

PR #1377, opened 2018-07-09
Develop

ezturner:


travis-ci/urls.py
```diff
@@ -6,4 +6,5 @@
     url(r'^logging', include('django_client_logger.urls')),
     url(r'^logging', include('rc_django.urls')),
     url(r'^', include('myuw.urls')),
+    url(r'^saml/', include('uw_saml.urls')),
```
fanglinfang: does this order matter?

ezturner: Good catch! I think it will

PR #1475, opened 2018-10-10
Adding catchall logging and exception handling

ezturner:
I believe we want to have a catchall exception handler here so that misformated messages get redirected to the dead letter queue rather than breaking the polling process.


fanglinfang:
Close this PR.
I will update section_status.py to include validating the message body. The changes will be in the python upgrade PR.

PR #1513, opened 2019-01-14
Bugfix/muwm 4475

ezturner:


myuw/dao/visual_schedule.py
```diff
@@ -698,16 +698,23 @@ def _add_dates_to_sections(schedule):
...
                 else:
                     section.start_date = schedule.term.first_day_quarter
+
+        if section.end_date is None:
+            if section.summer_term == "":
+                section.end_date = schedule.term.last_day_instruction
+            else:
+                if section.summer_term == "A-term":
+                    section.end_date = schedule.term.aterm_last_date
+                elif section.summer_term == "B-term":
```
fanglinfang: Would you check if line 715-716 can be eliminated?

myuw/dao/visual_schedule.py
```diff
@@ -698,16 +698,23 @@ def _add_dates_to_sections(schedule):
         if section.start_date is None:
             if section.summer_term == "":
                 section.start_date = schedule.term.first_day_quarter
-                section.end_date = schedule.term.last_day_instruction
             else:
                 if section.summer_term == "A-term":
```
fanglinfang: Perhaps this can be simplified to just check for summer-b term.

fanglinfang: can you eliminate line 703-704 and start 705 with if ?

PR #1533, opened 2019-03-25
Fix/muwm 4489

devights:


devights:
New behavior per Matt Winslow:  Respect TSPrint for non-independent study sections, ignore TSPrint on independent study.

PR #1550, opened 2019-05-16
Resolve MUWM-4489 - fix a bug in commit 3ea16cad8

fanglinfang:


myuw/dao/registration.py
```diff
@@ -2,6 +2,7 @@
 This module provides access to uw_sws registration module
 """
 
+from copy import deepcopy
```
ezturner: Why is deepcopy necessary here?

ezturner: Oh, because you're altering a list you're iterating through?

fanglinfang: Because removing the list item will mess up the list (some item may be skipped)

ezturner: I think you might want copy() there instead, since you aren't modifying the instructor objects, merely the list

PR #1555, opened 2019-06-05
Adding initialization to variables

ezturner:


ezturner:
Built on my-dev. Not initializing your index variables here puts them in the global context - and handlebars uses i in a for loop in their compile function, causing the for loop to not evaluate correctly.

PR #1578, opened 2019-07-25
Remove alpine link from UW Resources

ezturner:


ezturner:
Built on my-dev. Bug is https://jira.cac.washington.edu/browse/MUWM-4516

PR #1655, opened 2019-12-03
Add target_group in myuwnotice object

fanglinfang:
https://jira.cac.washington.edu/browse/MUWM-4530
Built on my-test.


myuw/dao/myuw_notice.py
```diff
@@ -15,6 +16,11 @@ def get_myuw_notices_for_user(request):
             # exclude those past display time window
             continue
 
+        if notice.has_target_group():
+            is_effective_member(request, notice.target_group)
```
devights: Notice is appended regardless of the is_effective_member check returning false

fanglinfang: Thank you, Stephen!  I fixed it.

PR #1672, opened 2020-01-07
Resolve MUWM-4482

fanglinfang:
https://jira.cac.washington.edu/browse/MUWM-4482

The cron job will run daily at 7pm. Will finish by 7am.

### DIFF
--- a/myuw/management/commands/clear_expired_sessions.py
+++ b/myuw/management/commands/clear_expired_sessions.py
@@ -3,9 +3,9 @@ The django clearsessions commend internally calls:
    cls.get_model_class().objects.filter(
        expire_date__lt=timezone.now()).delete()
 which could lock the DB table for a long time when
- having a large number of records to delete.
+having a large number of records to delete.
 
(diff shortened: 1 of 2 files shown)


myuw/management/commands/clear_expired_sessions.py
```diff
@@ -0,0 +1,31 @@
...
+
+logger = logging.getLogger(__name__)
+DEL_SIZE = 40000
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        i = 0
+        for record in Session.objects.filter(expire_date__lt=timezone.now()):
```
devights: What about using the built in QuerySet delete; that's supposed to do them in bulk?  
https://docs.djangoproject.com/en/3.0/topics/db/queries/#deleting-objects
The DEL_SIZE limit could still be achieved by limiting the results in the Session query

fanglinfang: Thank you! Updated.

PR #1884, opened 2020-08-04
MUWM-4741

fanglinfang:
https://jira.cac.washington.edu/browse/MUWM-4741


lmanes:
I still see it as plain text, not a link. Is this possibly related to the way the web service error faker works (Stephen says it is session-based)?

PR #1911, opened 2020-08-18
Vue/int styling

charlon:


myuw/static/vue/components/index/cards/notices.vue
```diff
@@ -111,6 +113,6 @@ export default {
 
 <style lang="scss" scoped>
 .myuw-notice-list {
```
SarthakSingh31, line 115: Is this class no longer needed? If so it should be removed

PR #1950, opened 2020-09-08
Vue/husky experience card

j0pp:


myuw/static/vue/components/cards/husky-experience.vue
```diff
@@ -0,0 +1,45 @@
...
+      article_html: (state) => state.hx_toolkit.value.article_html,
+      hxtViewer: (state) => state.user.affiliations.hxt_viewer,
+    }),
+    ...mapGetters('hx_toolkit', {
+      isReady: 'isReady',
+      isErrored: 'isErrored',
+    }),
+  },
+  created() {
+    this.fetch();
```
SarthakSingh31, line 36: the fetch function should be only be called when hxtViewer is true so that we don't fetch data that we don't need

myuw/static/vue/store/hx_toolkit.js
```diff
@@ -0,0 +1,24 @@
...
+    );
+
+    if(htmlDoc.links[0] !== undefined) {
+        data.article_html = htmlDoc.links[0].outerHTML;
+    }
+    
+    return data;
+}
+const customActions = {
+    fetch: fetchBuilder('/api/v1/hx_toolkit/week/', postProccess, 'text'),
```
SarthakSingh31, line 19: as done as https://github.com/uw-it-aca/myuw/blob/ba022783306e6b711dac11bb39a593f67ff9b16f/myuw/static/js/ws_data.js#L1347-L1376, week should be passed as a url param. Example https://github.com/uw-it-aca/myuw/blob/feature/vue/myuw/static/vue/components/cards/grades.vue#L188
and 
https://github.com/uw-it-aca/myuw/blob/feature/vue/myuw/static/vue/store/courses.js#L10

SarthakSingh31 (changes requested):

SarthakSingh31 (approved):

PR #2032, opened 2020-10-01
Unscope sass so card styling can be overridden.

charlon:


fanglinfang:
yield for myuw prod issue fix

SarthakSingh31 (approved):

PR #2054, opened 2020-10-08
Vue/restructure comps

charlon:
This is significant vue component restructuring as it was starting to get messy. This also changes the 'index' view to 'home' to have some consistency with what pages are referred to.


fanglinfang (commented):
There seems to be redundant directory levels:
1. It seems to be clearer to have "components/home_page" instead of "components/pages/home"? 
2. Could we eliminate the "cards" level in components/pages/home/?   Cards are the only containers under a page.

fanglinfang:
There seems to be redundant directory levels:
1. It seems to be clearer to have "components/home_page" instead of "components/pages/home"? 
2. Could we eliminate the "cards" level in components/pages/home/?   Cards are the only containers under a page.

fanglinfang (approved):

SarthakSingh31 (approved):

PR #2081, opened 2020-10-14
Created grad status card and added tests

j0pp:


myuw/static/vue/components/academics/grad-status.vue
```diff
@@ -0,0 +1,168 @@
+<template>
+  <uw-card
+    v-if="!isReady || is_grad"
+    :loaded="isReady" :errored="isErrored"
+    :errored-show="showError"
+  >
+    <template #card-heading>
```
fanglinfang, line 7: If the user has NO degrees, leaves, and petitions, we want to hide the card.

fanglinfang (approved):

PR #2107, opened 2020-10-30
Added textbook page

SarthakSingh31:


charlon (approved):

myuw/static/vue/components/textbooks/textbooks.vue
```diff
@@ -0,0 +1,156 @@
...
+    }),
+    ...mapGetters('textbooks', {
+      isTextbookReady: 'isReadyTagged',
+      isTextbookErrored: 'isErroredTagged',
+      statusCodeTextbooks: 'statusCodeTagged',
+      getProcessedData: 'getProcessedData',
+    }),
+    isReady() {
+      return (
+        this.isStudScheduleReady(this.term) &&
```
fanglinfang, line 112: What if this user is an instructor, not a student?

SarthakSingh31: https://github.com/uw-it-aca/myuw/blob/7f20bf56a3f3f3f1b791b3acb468e06a64fd14df/myuw/static/vue/components/textbooks/textbooks.vue#L114-L117

This value is false when the inst data is still fetching, so it wait for a response from the inst api.

fanglinfang: should we have is_student && isStudScheduleReady || is_instructor && isInstScheduleReady ?

SarthakSingh31: we need both of them to be fetched before loading the page, I will add a conditional on the isStudScheduleReady as well

fanglinfang (approved):

PR #2134, opened 2020-11-10
Add continuing education card

nick-falco:
Add the ContinuingEducation card to the home page


myuw/static/vue/components/home/former_student/ctnu-edu.vue
```diff
@@ -0,0 +1,42 @@
...
+import {mapState} from 'vuex';
+import Card from '../../_templates/card.vue';
+
+export default {
+  components: {
+    'uw-card': Card,
+  },
+  computed: {
+    ...mapState({
+      pastStudent: (state) => state.user.affiliations.past_stud,
```
fanglinfang, line 35: You want to map showCard: (state) => state.user.affiliations.no_1st_class_affi, 
and change to v-if="showCard" in line 3

fanglinfang (changes requested):

myuw/static/vue/components/home/former_student/ctnu-edu.vue
nick-falco: Oh I see, that makes sense. I was confused where the variable used in the v-if was coming from. Thanks!

fanglinfang (approved):

charlon (approved):
I added some additional styling classes and updated file names to be more explicit. Looks good!

PR #2163, opened 2020-11-25
F/muwm 4875

fanglinfang:


myuw/templates/home.html
```diff
@@ -41,7 +41,7 @@
       <myuw-retiree></myuw-retiree>
       <myuw-transcripts></myuw-transcripts>
       <myuw-continuing-education></myuw-continuing-education>
-      <myuw-uwnetid></myuw-uwnetid>
+      <myuw-uwnetid is-home-page=true></myuw-uwnetid>
```
SarthakSingh31, line 44: `is-home-page` is enough. If the attribute is present it is treated as true if it is missing it is treated as false

SarthakSingh31 (approved):

myuw/templates/home.html
fanglinfang: Ah, thanks!

PR #2171, opened 2020-12-08
F/class list

fanglinfang:


.travis.yml
```diff
@@ -54,7 +54,7 @@ after_success:
         "develop")
           export APP_INSTANCE="dev";
           ;;
-        "feature/vue")
+        "f/class-list")
```
SarthakSingh31, line 57: This probably should be changed back to `feature/vue`

myuw/static/vue/components/classlist/class-student-list.vue
```diff
@@ -0,0 +1,135 @@
+<template>
+  <uw-card v-if="showCard"
+           :loaded="isReady"
+           :errored="isErrored"
+           :errored-show="isErrored"
```
SarthakSingh31, line 5: This is not required as errored-show is true by default. The error will not be shown is errored is false.

.travis.yml
fanglinfang: Definitely, updated.

myuw/static/vue/components/classlist/class-student-list.vue
fanglinfang: Thanks, updated!

fanglinfang:
/app/myuw/static/vue/components/classlist/photo-list.vue

  13:1  error  This line has a length of 91. Maximum allowed is 80  max-len

PR #2185, opened 2020-12-25
F/profile

fanglinfang:


myuw/static/vue/components/accounts/medicine-account.vue
```diff
@@ -93,7 +92,7 @@ export default {
       return this.statusCode !== 404;
     },
     showCard() {
-      return !this.isReady || (this.isReady && this.hasActiveMedPw);
+      return !this.isReady || this.hasActiveMedPw || this.isErrored;
```
SarthakSingh31, line 95: `this.isErrored` is not required here, if the request errors out then this.isReady is still false.

fanglinfang: Ah, thanks!

SarthakSingh31 (approved):

PR #2334, opened 2021-04-01
Vue/resources style

j0pp:


fanglinfang:
Adding a clear separation between sections would be helpful.

fanglinfang:
When narrowing the window, there is a 3-column display:
image

charlon (approved):

PR #2346, opened 2021-04-12
Fang/merge develop

fanglinfang:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM acait/django-container:1.2.5 as pre-container
+FROM gcr.io/uwit-mci-axdd/django-container:1.3.1 as app-container
```
SarthakSingh31, line 1: this needs to be `pre-container`

Dockerfile
```diff
@@ -22,16 +22,14 @@ ARG VUE_DEVTOOLS
 ENV VUE_DEVTOOLS=$VUE_DEVTOOLS
 RUN npx webpack
 
-FROM pre-container as app-container
```
SarthakSingh31, line 25: this line is also required

Dockerfile
fanglinfang: Thanks!

SarthakSingh31 (approved):

PR #2356, opened 2021-04-19
hide make payment button if balance is 0

fanglinfang:


myuw/static/vue/components/accounts/tuition-fees.vue
```diff
@@ -48,7 +48,7 @@
                     </a>
                 </div>
               </div>
-              <div class="text-right">
+              <div v-if="tuiBalance != 0" class="text-right">
```
fanglinfang, line 51: show Make Payment link only when balance is not 0

SarthakSingh31 (approved):

PR #2438, opened 2021-06-15
minor fix

fanglinfang:


myuw/static/vue/components/profile/applicant-profile.vue
```diff
@@ -37,7 +39,7 @@
       <uw-card-property-group>
         <uw-card-property title="Email Address">
           <span v-if="email">{{ email }}</span>
-          <span v-else>No email address availabe</span>
+          <span v-else class="text-muted">No email address availabe</span>
```
charlon, line 42: Typo... should be "available"

fanglinfang: Ah, good catch!

PR #2498, opened 2021-07-16
Resolve MUWM-4981

fanglinfang:


myuw/static/vue/vuex/store/notices.js
```diff
@@ -61,6 +61,8 @@ const postProcess = (response, _, rootState) => {
   });
 };
 
+window.noticePostProcess = postProcess;
```
fanglinfang, line 64: Not sure why this is here

PR #2513, opened 2021-07-27
convert b-collapse to div

fanglinfang:


fanglinfang:
We are having the error: 'v-model' directives aren't supported on <div> elements

charlon (approved):

PR #3030, opened 2023-11-21
MUWM-5272

fanglinfang:
https://jira.cac.washington.edu/browse/MUWM-5272


myuw_vue/components/textbooks/textbooks.vue
```diff
@@ -1,6 +1,12 @@
 <template>
   <uw-panel :loaded="isReady" :errored="isErrored">
     <template #panel-body>
+      <div
+        v-if="student && !tacoma"
+        class="alert myuw-text-md" role="alert"
+      >
+        UW Day One Access Program
```
fanglinfang, line 8: William, please add the textbook static content here.

myuw_vue/components/accounts/tuition-fees.vue
```diff
@@ -130,6 +130,55 @@
             </template>
           </uw-card-status>
         </li>
+
+        <li v-if="hasIacData">
+          <h3>UW Day One Access Fees</h3>
+          <div>
+            One of ...
```
fanglinfang, line 137: William, please add the static content.

docker/settings.py
```diff
@@ -207,9 +207,11 @@
     }
 }
 
-DEBUG = False
+DEBUG = True
```
devights, line 210: Is this a permanent change? Any exceptions will include most of the Django settings and other configuration in the output to the user.  

Also, if debug is False ALLOWED_HOSTS must be set or all requests will fail: https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-DEBUG

fanglinfang: No, not permanent .

devights (approved):

PR #3161, opened 2024-09-27
Resolve MUWM-5364

fanglinfang:
https://jira.cac.washington.edu/browse/MUWM-5364

### DIFF
--- a/myuw_vue/components/home/summaries.vue
+++ b/myuw_vue/components/home/summaries.vue
@@ -47,7 +47,7 @@
         </span>
       </a>
     </div>
-    <div v-if="isHfsReady || isLibraryReady" class="col-md-10">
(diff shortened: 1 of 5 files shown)


myuw_vue/components/home/summaries.vue
```diff
@@ -177,10 +182,15 @@ export default {
       isLibraryReady: 'isReady',
       isLibraryErrored: 'isErrored',
     }),
+    loadData() {
+      return (this.alum || this.student || this.past_stud || this.employee || this.past_emp);
```
fanglinfang, line 186: Husky card only shows for student, past_stud, employee, or past_employee. We should match that on home page. This reduces unnecessary requests to HFS and Lib account apis.

